PR #7, opened 2015-10-16
Progress on fixing the builds

bootstraponline:
This fixes building the `selendroid-test-app` project. The main change is moving these to individual gradle projects instead of having a settings.gradle and build.gradle at the root. Without this change, everything gets built together which is what causes the build failures. We have desktop only jars and android only apks, they shouldn't be mixed.

The rest of the project will need to be updated with similar fixes.


vishals79:
Moved the changes to individual gradle projects with settings.gradle for projects wherever required.
All the projects are compiling fine but
1. Selendroid-test-app is dependent on Standalone,Client and Server-Common. Removing these project dependencies from build.gradle are causing build failures. After adding the dependencies and settings.gradle for selendroid-test-app we are getting exception. Please find the attached log file.
[Selendroid-test-app-build-log.txt](https://github.com/appium/appium-uiautomator2-driver/files/14790/Selendroid-test-app-build-log.txt)

bootstraponline:
> Removing these project dependencies from build.gradle are causing build failures.

How are you building selendroid-test-app? The following works for me:

```
$ cd selendroid-test-app
$ gradle clean assembleDebug
```

- https://gist.github.com/bootstraponline/20f554de6ff3a0b10d1a

vishals79:
@bootstraponline : selendroid-test-app is building fine. I have done changes for other projects too. Please review it once then we can merge this branch with the master.

bootstraponline:
@vishals79 Thanks. I have commented on the changes. Once fixed then this is ready to merge.

vishals79:
@bootstraponline : Thanks for the review. I have included the mentioned changes and merging the branch with master.

PR #11, opened 2015-10-19
Uiautomator v2 impl

bootstraponline:
When a branch is ready for merging / comments, then it's best to submit a PR

The commit history should be cleaned up. See the [rebasing section of appium's style guide](https://github.com/appium/appium/blob/c6269889e35626740a3ac9f74ce5ebbef348cae5/docs/en/contributing-to-appium/style-guide.md#rebasing). It's fine for now since this is a work in progress, however before changes are submitted upstream we'll want a clean history.

- The `Created by` lines need to be deleted. Git is used to track authorship information. Having people type their names into the Java files every time someone new edits the file doesn't scale.
- Dead code needs to be deleted, not left commented. Ex: `//import android.support.test.uiautomator.UiDevice;`


selendroid-server/src/main/java/io/selendroid/server/handler/UiAutomatorSendKeysAction.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public UiAutomatorSendKeysAction(String mappedUri) {
+        super(mappedUri);
+    }
+
+    @Override
+    public Response safeHandle(HttpRequest request) throws JSONException {
+        SelendroidLogger.info("send keys command using ui automator");
+        String[] keysToSend = extractKeysToSendFromPayload(request);
+        SelendroidDriver appiumUiAutomatorDriver=getSelendroidDriver(request);
```
bootstraponline: I recommend running the Android Studio formatter on the new code that's being written for appium. This will ensure there's consistency of style in the code. `appiumUiAutomatorDriver=getSelendroidDriver(request);` isn't good because there's normally a space before and after the `=`

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.Set;
+
+import io.selendroid.server.ServerInstrumentation;
+import io.selendroid.server.android.InstrumentedKeySender;
+import io.selendroid.server.android.KeySender;
+import io.selendroid.server.android.internal.Dimension;
+import io.selendroid.server.common.exceptions.SelendroidException;
+import io.selendroid.server.util.SelendroidLogger;
+
+//import android.support.test.uiautomator.UiDevice;
```
bootstraponline: Delete unused imports.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,288 @@
...
+    public static final String ACCEPT_SSL_CERTS = "acceptSslCerts";
+    public static final String SUPPORTS_NETWORK_CONNECTION = "networkConnectionEnabled";
+    private static UiDevice device;
+    private ServerInstrumentation serverInstrumentation = null;
+    private KeySender keySender = null;
+    private Session session = null;
+
+    public DefaultAppiumUiAutomatorDriver(ServerInstrumentation serverInstrumentation) {
+        this.keySender = new InstrumentedKeySender(serverInstrumentation);
+        ;
```
bootstraponline: `;` should be deleted.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,288 @@
...
+        try {
+            JSONObject capabilities = session.getCapabilities();
+            if (capabilities != null) {
+                copy = new JSONObject(capabilities.toString());
+            } else {
+                copy = new JSONObject();
+            }
+            copy.put(TAKES_SCREENSHOT, true);
+            copy.put(BROWSER_NAME, "selendroid");
+            copy.put("automationName", "UiAutomator");
```
bootstraponline: These string keys should be refactored to be `public static final String` constants just like `TAKES_SCREENSHOT`, `BROWSER_NAME`, and the others.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,288 @@
...
+
+    @Override
+    public void setAlertText(CharSequence... keysToSend) {
+
+
+    }
+
+    @Override
+    public void acceptAlert() {
+
```
bootstraponline: I think these should throw a not implemented exception so we know when they're called.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriverFactory.java
```diff
@@ -4,9 +4,24 @@
 
 public class DefaultSelendroidDriverFactory implements SelendroidDriverFactory {
 
+    private static SelendroidDriver selendroidDriver = null;
+
+    //How about having an enum?
+    private static final String AN_SELENDROID = "selendroid";
+    private static final String AN_UIAUTOMATOR = "UiAutomator";
```
bootstraponline: I think static Strings are fine. The `AN_` prefix should be removed.

```java
private static final String AN_SELENDROID = "selendroid";
private static final String AN_UIAUTOMATOR = "UiAutomator";
```

selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriverFactory.java
```diff
@@ -4,9 +4,24 @@
...
+
     @Override
-    public SelendroidDriver createSelendroidDriver( ServerInstrumentation androidInstrumentation ) {
-        return new DefaultSelendroidDriver(androidInstrumentation);
+    public SelendroidDriver createSelendroidDriver( ServerInstrumentation androidInstrumentation, String automationName ) {
+
+        if(selendroidDriver == null) {
```
bootstraponline: Formatting of this code is weird. Android Studio should auto format it correctly.

bootstraponline:
I think the goal should be to clean up what we have and then submit a PR to Selendroid so we can collect feedback. Otherwise we might end up deleting a bunch of work if they don't like the overall approach.

selendroid-server/AndroidManifest.xml
```diff
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="0.17.0-SNAPSHOT" package="io.selendroid.server">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:versionCode="1"
+    android:versionName="0.17.0-SNAPSHOT"
+    package="io.selendroid.server">
 
-    <uses-sdk android:minSdkVersion="10"/>
+    <uses-sdk android:minSdkVersion="18"/>
```
bootstraponline: Selendroid is intended to work on lower API levels so we can't just set the min API to 18.

selendroid-server/build.gradle
```diff
@@ -15,7 +15,7 @@ android {
     buildToolsVersion "23.0.1" // https://developer.android.com/tools/revisions/build-tools.html
 
     defaultConfig {
-        minSdkVersion 10
+        minSdkVersion 18
```
bootstraponline: This should remain at 10. Appium users rely on Selendroid for API 10+ automation.

selendroid-server/src/main/java/io/selendroid/server/ServerInstrumentation.java
```diff
@@ -136,6 +137,11 @@ public void onCreate(Bundle arguments) {
       extensionLoader = new ExtensionLoader(context);
     }
 
+    //set AutomationName
+    if(args.getAutomationName() != null){
```
bootstraponline: spacing is weird, normally it'd be like this: `if (args.getAutomationName() != null) {`

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+            SelendroidLogger.info("capabilities: " + copy);
+            return copy;
+        } catch (JSONException e) {
+            throw new SelendroidException(e);
+        }
+    }
+
+    @Override
+    public JSONObject getFullWindowTree() {
+        throw new NoSuchMethodImplementationException("Method not implemented");
```
bootstraponline: Rather than repeat the string every time, the exception should just default to using `"Method not implemented"` if the message isn't set explicitly.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    @Override
+    public String getWindowSource() {
+        throw new NoSuchMethodImplementationException("Method not implemented");
+    }
+
+    @Override
+    public String initializeSession(JSONObject desiredCapabilities) {
+        //Temporary solution for session creation and this needs to be modified.
```
bootstraponline: Selendroid uses a space after `//` so it should be: `// Temporary solution for session creation and this needs to be modified.`

selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriverFactory.java
```diff
@@ -4,9 +4,22 @@
...
 
+    private static SelendroidDriver selendroidDriver = null;
+    private static final String SELENDROID = "selendroid";
+    private static final String UIAUTOMATOR = "uiautomator";
+
     @Override
-    public SelendroidDriver createSelendroidDriver( ServerInstrumentation androidInstrumentation ) {
-        return new DefaultSelendroidDriver(androidInstrumentation);
+    public SelendroidDriver createSelendroidDriver(ServerInstrumentation androidInstrumentation, String automationName) {
+        if (selendroidDriver == null) {
```
bootstraponline: Why isn't this code just:

```java
if (automationName.equals(UIAUTOMATOR)) {
  return new DefaultAppiumUiAutomatorDriver(androidInstrumentation);
} else {
  return new DefaultSelendroidDriver(androidInstrumentation);
}
```

The singleton logic around selendroidDriver seems wrong since it's never assigned a value.

selendroid-server/src/test/java/io/selendroid/server/BaseTest.java
```diff
@@ -45,12 +45,13 @@
   public static final String host = "127.0.0.1";
   protected AndroidServer server;
   public static final String ANY_STRING = "ANY-STRING";
+  private static String automationName = "selendroid";
```
bootstraponline: This should be final and in caps.

`private static final String AUTOMATION_NAME = "selendroid";`

selendroid-server/src/test/java/io/selendroid/server/BaseTest.java
```diff
@@ -45,12 +45,13 @@
...
 
   @Before
   public void setup() throws Exception {
     ServerInstrumentation instrumentation = mock(ServerInstrumentation.class);
     when(instrumentation.getServerVersion()).thenReturn("0.2");
-    server = new AndroidServer(instrumentation, port);
+    server = new AndroidServer(instrumentation, port,automationName);
```
bootstraponline: spacing:
`server = new AndroidServer(instrumentation, port, automationName);`

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -236,6 +236,11 @@ public void startSelendroid(AndroidApp aut, int port, SelendroidCapabilities cap
         argList.addAll(Lists.newArrayList("-e", "bootstrap", capabilities.getBootstrapClassNames()));
       }
     }
+
+    if(capabilities.getAutomationName() != null){
```
bootstraponline: spacing:

`if (capabilities.getAutomationName() != null) {`

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    @Override
+    public void refresh() {
+        throw new NoSuchMethodImplementationException("Method not implemented");
+    }
+
+    @Override
+    public boolean isAlertPresent() {
+        return false;
```
bootstraponline: this is not a correct implementation of the method, selendroid has actual logic that detects alert presence.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+    @Override
+    public Session getSession() {
+        return session;
+    }
+
+    @Override
+    public JSONObject getSessionCapabilities(String sessionId) {
+        SelendroidLogger.info("session: " + sessionId);
+
+        JSONObject copy;
```
moizjv: what does copy mean? it is very ambiguous? Can we have a better name?

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+        try {
+            JSONObject capabilities = session.getCapabilities();
+            if (capabilities != null) {
+                copy = new JSONObject(capabilities.toString());
+            } else {
+                copy = new JSONObject();
+            }
+            copy.put(TAKES_SCREENSHOT, true);
+            copy.put(BROWSER_NAME, "selendroid");
```
moizjv: just wondering; what if capabilities object had these set already; do we want to reset all capabilities?

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,279 @@
...
+import io.selendroid.server.android.KeySender;
+import io.selendroid.server.android.internal.Dimension;
+import io.selendroid.server.common.exceptions.NoSuchMethodImplementationException;
+import io.selendroid.server.common.exceptions.SelendroidException;
+import io.selendroid.server.util.SelendroidLogger;
+
+
+public class DefaultAppiumUiAutomatorDriver implements SelendroidDriver {
+
+    public static final String BROWSER_NAME = "browserName";
```
moizjv: why is this public static; will these constants be used out side this context; is there better mechanism for declaring constants

selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriverFactory.java
```diff
@@ -4,9 +4,22 @@
...
+            if (automationName.equals(SELENDROID)) {
+                return new DefaultSelendroidDriver(androidInstrumentation);
+            } else if (automationName.equals(UIAUTOMATOR)) {
+                //custom appium Uiautomatorv2 driver
```
moizjv: space after `// custom` ..

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
PrakashGoulla: The default capabilities will be set only if the session does not contain any capabilities.
```
     if (capabilities != null) {
                copy = new JSONObject(capabilities.toString());
            } else {
                copy = new JSONObject();
            }
```
and yes the var name `copy` needs to changed.

selendroid-server/src/main/java/io/selendroid/server/model/DefaultAppiumUiAutomatorDriver.java
```diff
@@ -0,0 +1,286 @@
...
+    public static final String TAKES_SCREENSHOT = "takesScreenshot";
+    public static final String SUPPORTS_ALERTS = "handlesAlerts";
+    public static final String ROTATABLE = "rotatable";
+    public static final String ACCEPT_SSL_CERTS = "acceptSslCerts";
+    public static final String SUPPORTS_NETWORK_CONNECTION = "networkConnectionEnabled";
+    public static final String AUTOMATION_NAME = "automationName";
+    public static final String PLATFORM_NAME = "platformName";
+    public static final String PLATFORM_VERSION = "platformVersion";
+    public static final String SERVER_VERSION = "version";
+    public static final String METHOD_NOT_IMPLMENTED_EXCEPTION = "Method has no Implementation";
```
moizjv: does this need to be `public static`?

anandsadu:
https://github.com/anandsadu/appium-uiautomator-selendroid-integration.git

anandsadu:
Code moved to https://github.com/anandsadu/appium-uiautomator-selendroid-integration.git.
Using this repo for the uiautomator2 driver implementation with  Appium 1.5

PR #16, opened 2015-10-22
Rebase on upstream

bootstraponline:
selendroid released 0.17.0 so it's important we stay up to date with upstream changes.


bootstraponline:
selendroid released 0.18.0 so it's important we stay up to date with upstream changes.

Issue #8, opened 2015-10-19
Publish release jars to jcenter & maven central


bootstraponline:
Ensure our package names don't conflict with the upstream selendroid jars.

- [droiddriver jcenter.gradle](https://github.com/appium/droiddriver/blob/72a40802f8e7d24a3175e09adc4b44a89df54f38/jcenter.gradle)

bootstraponline:
Migrated to https://github.com/appium/selendroid/issues/2

PR #23, opened 2016-06-09
Added MultiTouch Functionality

sravanmedarapu:
- Implemented by taking the reference from [appium-android-driver](https://github.com/appium/appium-android-driver/blob/master/lib/commands/touch.js) module implementation

- MultiPointerGesture Handler(`/touch/multi/perform`) which will be invoked from [`performMultiAction`](https://github.com/sravanmedarapu/appium-uiautomator2-driver/blob/2791e11e5a0aa22c4fda3ac3c9bfff7aed71133f/lib/commands/touch.js#L120),  implemented in UiAutomator2-server module - [#8](https://github.com/appium/appium-uiautomator2-server/pull/8)

### DIFF
--- a/lib/commands/touch.js
+++ b/lib/commands/touch.js
@@ -1,4 +1,3 @@
-import log from '../logger';
 import _ from 'lodash';
 import androidHelpers from '../android-helpers';
 import { asyncmap } from 'asyncbox';
@@ -23,7 +22,7 @@ helpers.parseTouch = async function (gestures, multi) {
           options.x = pos.x + (gesture.options.x || 0);
           options.y = pos.y + (gesture.options.y || 0);
         } else {
-          options.x =  pos.x + (size.width / 2);
+          options.x = pos.x + (size.width / 2);
           options.y = pos.y + (size.height / 2);
         }
         let touchStateObject = {
@@ -94,11 +93,11 @@ helpers.parseTouch = async function (gestures, multi) {
 
 
 commands.performMultiAction = async function (actions, elementId) {
- 
+
   // Android needs at least two actions to be able to perform a multi pointer gesture
   if (actions.length === 1) {
     throw new Error("Multi Pointer Gestures need at least two actions. " +
-                    "Use Touch Actions for a single action.");
+        "Use Touch Actions for a single action.");
   }
 
   let states = await asyncmap(actions, async (action) => {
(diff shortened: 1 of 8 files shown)


lib/commands/touch.js
```diff
@@ -0,0 +1,124 @@
...
+
+let commands = {}, helpers = {}, extensions = {};
+
+helpers.parseTouch = async function (gestures, multi) {
+  // because multi-touch releases at the end by default
+  if (multi && _.last(gestures).action === 'release') {
+    gestures.pop();
+  }
+
+  let touchStateObjects = await asyncmap(gestures, async (gesture) => {
```
imurchie: Does this pass linting? JSHint has problems, as far as I know, with async functions using the fat arrow.

lib/driver.js
```diff
@@ -23,9 +23,11 @@ const SYSTEM_PORT_RANGE = [8200, 8299];
 // one of the ports above on the system to this port on the device.
 const DEVICE_PORT = 8080;
 
-// NO_PROXY contains the paths that we never want to proxy to UiAutomator2.
-// TODO  Add the list of paths that we never want to proxy to UiAutomator2.
-const NO_PROXY = [];
+// NO_PROXY contains the paths that we never want to proxy to UiAutomator2 server.
+// TODO  Add the list of paths that we never want to proxy to UiAutomator2 server.
+const NO_PROXY = [
+['POST', new RegExp('^/session/[^/]+/touch/multi/perform')]
```
imurchie: This should be indented.

imurchie:
A minor issue and a question. Otherwise looks good!

sravanmedarapu:
Refactored code to address indention issue wherever required.

lib/commands/touch.js
```diff
@@ -0,0 +1,124 @@
+import _ from 'lodash';
+import androidHelpers from '../android-helpers';
+import { asyncmap } from 'asyncbox';
+
+let commands = {}, helpers = {}, extensions = {};
+
+helpers.parseTouch = async function (gestures, multi) {
```
jlipps: if this logic is the same as from appium-android-driver, we should definitely export it there and import it here rather than duplicating it.

sravanmedarapu: Not exactly same, we are proxying the request to UiAutomator2 server here, where as in appium-android-driver we are processing using bootstrap.

PR #25, opened 2016-06-21
Added swipe action and integrated with server module with specs

anandsadu:
Added swipe action in the driver module and integrated with server module.Added specs for swipe.

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -1,0 +1,28 @@
+import androidHelpers from '../android-helpers';
+import log from '../logger';
+
+const swipeStepsPerSec = 28;
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.swipe = async function (startX, startY, endX, endY, duration, touchCount, elId) {
+  if (startX === 'null') {
+    startX = 0.5;
+  }
+  if (startY === 'null') {
+    startY = 0.5;
+  }
+  let swipeOpts = {startX, startY, endX, endY,
+                   steps: Math.round(duration * swipeStepsPerSec)};
+
+  if (typeof elId !== "undefined" && elId !== null) {
+    swipeOpts.elementId = elId;
+    return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
+  } else {
+    return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
+  }
+};
+
+Object.assign(extensions, commands, helpers);
+export { commands, helpers };
+export default extensions;
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -16,10 +16,8 @@ commands.swipe = async function (startX, startY, endX, endY, duration, touchCoun
 
   if (util.hasValue(elId)) {
     swipeOpts.elementId = elId;
-    return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
-  } else {
-    return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
-  }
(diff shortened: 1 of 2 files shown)


lib/commands/actions.js
```diff
@@ -0,0 +1,28 @@
+import androidHelpers from '../android-helpers';
+import log from '../logger';
+
+const swipeStepsPerSec = 28;
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.swipe = async function (startX, startY, endX, endY, duration, touchCount, elId) {
+  if (startX === 'null') {
```
imurchie: Should this check for both `null` and `undefined`? If you use [utils](https://github.com/appium/appium-support/blob/master/lib/util.js) from `appium-support` there is a method `hasValue` that checks.

```
import { util } from 'appium-support';

...

startX = util.hasValue(startX) ? startX : 0.5;
startY = util.hasValue(startY) ? startY : 0.5;
```

test/functional/commands/touch/swipe-e2e-specs.js
```diff
@@ -0,0 +1,28 @@
...
+describe('apidemo - touch - swipe-action', function () {
+  before(async () => {
+    driver = new AndroidUiautomator2Driver();
+    await driver.createSession(defaultCaps);
+  });
+  after(async () => {
+    await driver.deleteSession();
+  });
+  it('should swipe', async () => {
+  await driver.swipe(100,650,100,330,1000);
```
imurchie: Is there nothing that can be asserted on this? Perhaps the visible elements are different?

lib/commands/actions.js
```diff
@@ -0,0 +1,28 @@
...
+    startX = 0.5;
+  }
+  if (startY === 'null') {
+    startY = 0.5;
+  }
+  let swipeOpts = {startX, startY, endX, endY,
+                   steps: Math.round(duration * swipeStepsPerSec)};
+
+  if (typeof elId !== "undefined" && elId !== null) {
```
imurchie: Again, `hasValue` would be useful here. It is also good to just add the element id to the opts, and then do the command outside of the conditional, so the call is not repeated.

```js
if (util.hasValue(elId)) {
  swipeOpts.elementId = elId;
}
return await this.uiautomator2.jwproxy.command(`/touch/perform`, 'POST', {swipeOpts});
```

test/functional/commands/touch/swipe-e2e-specs.js
anandsadu: Used coordinates because swipe is performed on the device screen and not on elements.One option could be perform the action between two elements.

imurchie: Understood. But there is no way, without watching it, that one could tell that the swipe happened or did not happen.

lib/commands/actions.js
```diff
@@ -0,0 +1,27 @@
...
+ 
+ startX = util.hasValue(startX) ? startX : 0.5;
+ startY = util.hasValue(startY) ? startY : 0.5;
+
+  let swipeOpts = {startX, startY, endX, endY,
+                   steps: Math.round(duration * swipeStepsPerSec)};
+
+  if (util.hasValue(elId)) {
+    swipeOpts.elementId = elId;
+    return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
```
imurchie: There is still no reason to have the call to `command` twice. Once the `elementId` is in the `swipeOpts` the call is the same, no?

test/functional/commands/touch/swipe-e2e-specs.js
anandsadu: Sure. When find.js was checked in. I'll add the assert for it.

PR #26, opened 2016-06-22
Specs for driver module

sravanmedarapu:
- Added unit and functional specs for driver module.

- Reference [appium-android-driver](https://github.com/appium/appium-android-driver/blob/master/test/functional/commands/touch/multi-action-e2e-specs.js)   and [appium-selendroid-driver](https://github.com/appium/appium-selendroid-driver/blob/1f5b033f0ccd71efeff4ea717e037e6cb50e8b2a/test/unit/driver-specs.js)


lib/commands/element.js
```diff
@@ -36,5 +36,31 @@ commands.touchMove = async function (elementId, x, y) {
...
-Object.assign(extensions, commands);
+
+helpers.setElementValue = async function (keys, elementId, replace = false) {
+  let text = keys;
+  if (keys instanceof Array) {
+    text = keys.join("");
+  }
+
+  let params = {
+    "id" : elementId,
```
imurchie: Why is `id` in quotes?

imurchie:
How much of this code is copied and pasted from the android driver, with the command changed? This seems bad. It would be much better to figure out a way to share the code, abstracting out how the call is made to the underlying system. Otherwise we are going to get into the position of needing to update multiple places when something needs to be fixed or changed.

lib/commands/element.js
sravanmedarapu: As [SendKeysToElement ](https://github.com/appium/appium-uiautomator2-server/blob/master/app/src/main/java/io/appium/uiautomator2/handler/SendKeysToElement.java#L32) handler in uiautomator2-server module expects `id `in json body, just to be in sync with handler customized json body

sravanmedarapu:
@imurchie makes sense, Will try to investigate to figure out a way to do that. But seems there is no way or at least from my understanding to do that with out tweaking appium-android-driver module.

jlipps:
@sravanmedarapu it's ok to make a PR to the appium-android-driver module so that it exports tests or shared logic which can be used here. We've done that with basedriver already (it exports tests which the other drivers can import and run to ensure basic driver functionality works)

sravanmedarapu:
sure @jlipps, will try to export shared logic wherever required and make use of them here in uiautomator2-driver.

imurchie:
The Selendroid driver also imports bits of the Android driver. It doesn't currently need to modify any behavior (like sending to a different bootstrap module) but that could be accommodated.

See, for instance, https://github.com/appium/appium-selendroid-driver/blob/master/lib/driver.js#L11

lib/commands/element.js
imurchie: I just mean, the quotes are not necessary. `{ id: elementId }` is fine.

imurchie:
This looks good to me! It will have to wait until the android driver stuff gets merged, though.

PR #30, opened 2016-07-29
Added Webview support, tap, click, drag functions and wrapBootstrapDisconnect for network actions

sravanmedarapu:
- Added drag, tap and click

- Handled webview support, added exclusive `CHROME_NO_PROXY` list and also implemented `suspendChromedriverProxy` . 

- Added `killUiAutomatorOnDevice`  to kill the existing the running server instance if any on device.

- Added `wrapBootstrapDisconnect` to support network actions.

### DIFF
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -1,12 +1,10 @@
-import { util } from 'appium-support';
 
-const swipeStepsPerSec = 28;
 
 let commands = {}, helpers = {}, extensions = {};
 
 commands.doSwipe = async function(swipeOpts) {
   return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
-}
+};
 
 Object.assign(extensions, commands, helpers);
 export { commands, helpers };
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -6,6 +6,10 @@ commands.doSwipe = async function(swipeOpts) {
   return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
 };
 
+commands.doDrag = async function (dragOpts) {
+  return await this.uiautomator2.jwproxy.command(`/touch/drag`,'POST', dragOpts);
+};
+
 Object.assign(extensions, commands, helpers);
 export { commands, helpers };
 export default extensions;
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -3,7 +3,7 @@
 let commands = {}, helpers = {}, extensions = {};
 
 commands.doSwipe = async function(swipeOpts) {
-  return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
+  return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST', swipeOpts);
 };
 
 commands.doDrag = async function (dragOpts) {
--- a/lib/commands/actions.js
+++ b/lib/commands/actions.js
@@ -1,5 +1,3 @@
-import androidHelpers from '../android-helpers';
-import log from '../logger';
 import { util } from 'appium-support';
 
 const swipeStepsPerSec = 28;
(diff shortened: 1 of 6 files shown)


lib/commands/actions.js
```diff
@@ -1,14 +1,14 @@
...
 
 let commands = {}, helpers = {}, extensions = {};
 
 commands.doSwipe = async function(swipeOpts) {
   return await this.uiautomator2.jwproxy.command(`/touch/perform`,'POST',{swipeOpts});
-}
+};
+
+commands.doDrag = async function (dragOpts) {
+  return await this.uiautomator2.jwproxy.command(`/touch/drag`,'POST', dragOpts);
```
imurchie: Are we passing in something different here than in `doSwipe`? It seems strange to have the `swipeOpts` put into an object, while `dragOpts` is not. Normalizing things would be good.

lib/commands/element.js
```diff
@@ -40,6 +40,26 @@ commands.getText = async function (elementId) {
...
+    if (elementId) {
+      // we are either tapping on the default location of the element
+      // or an offset from the top left corner
+      if (x !== 0 || y !== 0) {
+        await this.uiautomator2.jwproxy.command(`/element/${elementId}/click`, 'POST', {elementId, x, y});
+      } else {
+        await this.uiautomator2.jwproxy.command(`/element/${elementId}/click`, 'POST', {elementId});
+      }
+    } else {
+      await this.uiautomator2.jwproxy.command(`/element/${elementId}/click`, 'POST', {x, y});
```
imurchie: In this block there is no `elementId` (or, at least, it is falsy), so what is put into the endpoint of the call? It could be any number of things, none of which make much sense (e.g., `'/element/undefined/click'`).

lib/commands/element.js
```diff
@@ -40,6 +40,26 @@ commands.getText = async function (elementId) {
...
+commands.click = async function (elementId) {
+  return await this.uiautomator2.jwproxy.command(`/element/${elementId}/click`, 'POST', {elementId});
+};
+
+commands.tap = async function (elementId, x = 0, y = 0, count = 1) {
+  for (let i = 0; i < count; i++) {
+    if (elementId) {
+      // we are either tapping on the default location of the element
+      // or an offset from the top left corner
+      if (x !== 0 || y !== 0) {
```
imurchie: I think the repetition of the actual command is bad. It would be better to get the correct parameters, and then send it.

```js
let params = {elementId};
if (x !== 0 || y !== 0) {
  params[x] = x;
  params[y] = y;
}
await this.uiautomator2.jwproxy.command(`/element/${elementId}/click`, 'POST', params);
```

This could be extended to include the other call, as well.

imurchie:
A couple of comments, but looks good!

lib/commands/element.js
sravanmedarapu: Right now click handler called with `/element/undefined/click` and server side we are relaying on jsonbody.  

In this case we may need to write another end point from server side like `/session/:sessionId/click` or   `/session/:sessionId/touch/click` whichever suitable best.

imurchie: Alternatively they could _all_ go to `/session/:sessionId/click`, since the element id is passed in as a parameter, too.

sravanmedarapu: Was referring to [JSONWireProtocol ](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidelementidclick) specs while implementing end points, but no issues will change to `/session/:sessionId/click` if fits best for our case.

imurchie: Yes. It would be best to follow the spec. Unfortunately the bare `/session/:sessionId/click` is for mouse buttons. 

There is [this command](https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidtouchdown).

sravanmedarapu: what do you think about adding custom one, something like `/session/:sessionId/appium/tap` and will take `x,y` coordinates as POST body i.e. `{ "x" : double, "y": double }`

imurchie:
LGTM.

PR #32, opened 2016-08-29
Refactored - download server apks while installing module and capabilites handling

sravanmedarapu:
- Refactored driver.js to `checkAppPresent` only on `app` cabability

- Downloading UiAutomator2 server apks while installing driver module


imurchie:
This all looks reasonable to me!

PR #34, opened 2016-09-13
Added Oreientation, Attribute related actions and specs

sravanmedarapu:
-  Updated NO_PROXY list and capabilities for the current session metadata

- Added attribute and Orientation related actions

- Added specs

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -1,44 +1,17 @@
 import log from '../logger';
-import utf7 from 'utf7';
 import androidHelpers from '../android-helpers';
-
-const {imap} = utf7;
 
 let extensions = {},
     commands = {},
     helpers = {};
-
-function encodeString (value, unicode) {
-  for (let i = 0; i < value.length; i++) {
-    let c = value.charCodeAt(i);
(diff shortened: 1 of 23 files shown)


lib/commands/general.js
```diff
@@ -1,40 +1,17 @@
...
+  return await this.uiautomator2.jwproxy.command('/source', 'GET', {});
+};
 
 // Need to override this for correct unicode support
-commands.keys = async function (value) {
-  if (value instanceof Array) {
-    value = value.join("");
-  }
-  log.debug(`Setting text: '${value}'`);
-  value = encodeString(value, this.opts.unicodeKeyboard);
```
imurchie: Does this system not support Unicode? Or does it actually natively support it so we can do away with the ime stuff?

imurchie:
Looks good.

lib/commands/general.js
sravanmedarapu: Tried in multiple versions of devices and observed it has native support.

imurchie: That is awesome!

PR #35, opened 2016-09-28
Addressed linting issues

sravanmedarapu:


.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 build
 .idea/
 uiautomator2/
+emulator.log
```
imurchie: You might want to make this just `*.log`.

package.json
```diff
@@ -31,29 +31,40 @@
...
+    "utf7": "^1.0.0",
+    "yargs": "^3.29.0"
   },
   "scripts": {
     "install": "node ./bin/install.js",
     "prepublish": "./node_modules/.bin/gulp prepublish",
     "test": "./node_modules/.bin/gulp once",
-    "watch": "./node_modules/.bin/gulp"
+    "watch": "./node_modules/.bin/gulp",
+    "lint": "gulp eslint"
```
imurchie: I personally like to have a message printed out here, to make it clear what is happening in pre-commit.

Like https://github.com/appium/appium-xcuitest-driver/blob/master/package.json#L53 and https://github.com/appium/appium-xcuitest-driver/blob/master/package.json#L57

sravanmedarapu: Sure, Will update.

PR #36, opened 2016-09-30
Changed device port number

sravanmedarapu:
depends on #https://github.com/appium/appium-uiautomator2-server/pull/26


imurchie:
👍

PR #40, opened 2016-10-19
Added specs and refactored logs

sravanmedarapu:


lib/uiautomator2.js
```diff
@@ -149,8 +149,8 @@ class UiAutomator2Server {
   async killUiAutomatorOnDevice () {
     try {
       await this.adb.killProcessesByName('io.appium.uiautomator2.server');
-    } catch (e) {
-      logger.warn(`Error while killing uiAutomator: ${e}`);
+    } catch (ignore) {
+      logger.info("Unable to kill the io.appium.uiautomator2.server process, assuming it is already killed");
```
imurchie: Is the error here not useful? Or is it always the same or something?

PR #45, opened 2016-10-27
Added SHA-2 hash check for server apks while downloading & added installer specs

sravanmedarapu:


lib/installer.js
```diff
@@ -28,19 +35,58 @@ async function setupUiAutomator2 () {
   }
 }
 
+async function fileExist (fileName, SHA512) {
```
imurchie: This function name does not really reflect the functionality, does it?

lib/installer.js
```diff
@@ -28,19 +35,58 @@ async function setupUiAutomator2 () {
...
+  if (await fs.exists(fileName)) {
+    var buffer = require("fs").readFileSync(fileName);
+    let fingerprint =  await sha512(buffer);
+    if ( fingerprint === SHA512 ) {
```
imurchie: This whole block is the same as saying `return fingerprint === SHA512`.

lib/installer.js
```diff
@@ -28,19 +35,58 @@ async function setupUiAutomator2 () {
...
+    throw new Error(Object.prototype.toString.call(serverApk));
+  }
+  if (!serverTestApk instanceof Buffer ) {
+    throw new Error(Object.prototype.toString.call(serverTestApk));
+  }
+  let serverFingerprint = await sha512(serverApk);
+  let serverTestFingerprint = await sha512(serverTestApk);
 
+  if ( serverFingerprint !== SERVER_DOWNLOAD_SHA512 ) {
+    log.errorAndThrow("bad Server SHA512 fingerprint: " + serverFingerprint );
```
imurchie: String template literal here.

lib/installer.js
```diff
@@ -28,19 +35,58 @@ async function setupUiAutomator2 () {
...
+    log.error("Stopping the installation" );
+    return;
+  }
+  if ( serverTestFingerprint !== SERVER_TEST_DOWNLOAD_SHA512 ){
+    log.errorAndThrow("bad Server test SHA512 fingerprint: " + serverTestFingerprint );
```
imurchie: String template literal here too.

PR #49, opened 2016-11-23
Updated ui2 server port forwarding approach

sravanmedarapu:
Fix  #https://github.com/appium/appium/issues/7270

Depends on #https://github.com/appium/appium-adb/pull/184

cc: @imurchie , @jlipps


lib/driver.js
```diff
@@ -10,6 +10,7 @@ import * as uiautomator2Helpers from './helpers';
 import { androidHelpers, androidCommands, WEBVIEW_BASE } from 'appium-android-driver';
 import desiredCapConstraints from './desired-caps';
 import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
+import {findAPortNotInUse} from 'portscanner';
```
jlipps: we do spaces around import braces, see the examples above this line

jlipps:
published as 0.2.0

PR #50, opened 2016-11-24
ensure we get latest server

jlipps:


jlipps:
@sravanmedarapu going ahead and merging this. if we actually don't want to auto-upgrade minor versions of the server, we can revert.

PR #55, opened 2016-12-09
Grant permissions and ignore exception from removePortForward

sravanmedarapu:
- Ignore if removePortForward throws error

- Call to grant permissions

https://github.com/appium/appium-adb/pull/189

@imurchie and @jlipps

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -282,6 +282,13 @@ class AndroidUiautomator2Driver extends BaseDriver {
     if (this.opts.app) {
       await helpers.installApkRemotely(this.adb, this.opts);
     }
+    if (this.opts.autoGrantPermissions) {
+      try {
+        await this.adb.grantAllPermissions(this.opts.appPackage, this.opts.app);
+      } catch (error) {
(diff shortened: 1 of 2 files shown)


jlipps:
@sravanmedarapu is there an issue that has reported this error? just want to link it if so

sravanmedarapu:
@jlipps No issue reported, felt adding a catch would be more appropriate. even failing `removePortForward` will not effect the end user.

jlipps:
sounds good

lib/driver.js
```diff
@@ -282,6 +282,13 @@ class AndroidUiautomator2Driver extends BaseDriver {
     if (this.opts.app) {
       await helpers.installApkRemotely(this.adb, this.opts);
     }
+    if (this.opts.autoGrantPermissions) {
+      try {
+        this.adb.grantAllPermissions(this.opts.appPackage, this.opts.app);
```
jlipps: shouldn't this be `await`ed?

jlipps:
don't forget to publish :-)

sravanmedarapu:
Sure :)

sravanmedarapu:
published in 0.2.2

PR #57, opened 2016-12-26
Refactor

sravanmedarapu:
- Updated

    - to use force stop in `killUiAutomatorOnDevice()`
    - ui2 server version and SHA hash to `v0.1.0`
    - regex to include only `session/:sessionId`


imurchie:
This looks fine to me.

PR #61, opened 2017-01-17
added device screen size, model and manufacturer details to session

sravanmedarapu:
Depends on #https://github.com/appium/appium-adb/pull/195

Req Ticket: #https://github.com/appium/appium/issues/3210

cc: @jlipps, @imurchie


imurchie:
The adb PR is closed, so the dependency could be bumped here and then this will be good to go!

PR #70, opened 2017-04-26
Added skipUnlock

mstrelex:
This will allow driver successful initialization even when the device screen remains locked.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -243,8 +243,12 @@ class AndroidUiautomator2Driver extends BaseDriver {
     logger.debug(`Forwarding UiAutomator2 Server port ${DEVICE_PORT} to ${this.opts.systemPort}`);
     await this.adb.forwardPort(this.opts.systemPort, DEVICE_PORT);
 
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -243,8 +243,12 @@ class AndroidUiautomator2Driver extends BaseDriver {
     logger.debug(`Forwarding UiAutomator2 Server port ${DEVICE_PORT} to ${this.opts.systemPort}`);
     await this.adb.forwardPort(this.opts.systemPort, DEVICE_PORT);
 
-     // unlock the device to prepare it for testing
-    await helpers.unlock(this, this.adb, this.caps);
+    if (!this.opts.skipUnlock) {
+      // unlock the device to prepare it for testing
+      await helpers.unlock(this, this.adb, this.caps);
+    } else {
+      logger.debug(`Skipping device unlock as instructed`);
```
imurchie: This log line is a bit vague. I'd rather something like `'skipUnlock' capability set, so skipping device unlock`

PR #71, opened 2017-04-27
Allow starting a session without providing an APK or package/activity

mstrelex:


imurchie:
Can you run the linter and fix the issues?

```
gulp eslint
```

imurchie:
Isn't all the `skipUnlock` stuff already in under a different PR? Why is it here now?

PR #72, opened 2017-04-28
Allow session initialization without APK or package/activity

mstrelex:


lib/driver.js
```diff
@@ -256,15 +259,21 @@ class AndroidUiautomator2Driver extends BaseDriver {
       await this.initAUT();
     }
     //Adding AUT package name in the capabilities if package name not exist in caps
-    if (!this.caps.appPackage) {
-      this.caps.appPackage = appInfo.appPackage;
+    if (this.opts.appInfo && !this.caps.appPackage) {
```
jlipps: this is not going to work. no `appInfo` property will ever exist on `this.opts`. The line above (`Object.assign(this.opts, appInfo);`) does not create a property called `appInfo`. It puts all the properties that exist on the `appInfo` object into `this.opts`. If you read the implementation of `helpers.getLaunchInfo`, you'll see that the last line is:

```js
return {appPackage, appWaitPackage, appActivity, appWaitActivity};
```

which means these are the properties that get updated on `this.opts` as a result of the `Object.assign`.

mstrelex: Thanks @jlipps, I missed this out... I will have a deeper look into that, meanwhile I will close the PR.

jlipps: sounds good! thanks for your idea and we'll be happy to assist when you've got another PR ready to look at.

PR #73, opened 2017-05-03
Make the timeout for the server to come up be configurable

imurchie:
Sometimes things are slow and the server does not start responding fast enough. 

See https://travis-ci.org/appium/appium-uiautomator2-driver/jobs/228376321#L3195-L3227

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -4,8 +4,11 @@ import { retryInterval } from 'asyncbox';
 import logger from './logger';
 import { UI2_SERVER_APK_PATH as apkPath, UI2_TEST_APK_PATH as testApkPath, UI2_VER} from './installer';
 import Promise from 'bluebird';
+import _ from 'lodash';
+
 
 const REQD_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort'];
+const SERVER_LAUNCH_RETRIES = 20;
 
 var adbkit = require('adbkit');
 var client = adbkit.createClient();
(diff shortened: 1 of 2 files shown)


lib/uiautomator2.js
```diff
@@ -119,9 +120,15 @@ class UiAutomator2Server {
       this.startSessionUsingAdbkit();
     }
 
-    logger.info('Waiting for UiAutomator2 to be online...');
+    let retries = SERVER_LAUNCH_RETRIES;
+    if (caps.uiautomator2ServerLaunchTimeout) {
+      caps.uiautomator2ServerLaunchTimeout = parseInt(caps.uiautomator2ServerLaunchTimeout, 10);
```
jlipps: isn't this already guaranteed to be a number? is there a reason to parse it?

lib/uiautomator2.js
```diff
@@ -119,9 +120,15 @@ class UiAutomator2Server {
...
+      retries = Math.round(caps.uiautomator2ServerLaunchTimeout / 1000);
+    }
+
+    logger.info(`Waiting up to ${retries * 1000}ms for UiAutomator2 to be online...`);
     // wait 20s for UiAutomator2 to be online
-    await retryInterval(20, 1000, async () => {
+    await retryInterval(retries, 1000, async () => {
```
jlipps: `retryInterval` is pretty smart. you can actually do this:

```js
await retryInterval(retries, 1000, this.jwproxy.command, '/status', 'GET');
```

to make it a one-liner! (assuming binding stuff is a-ok)

imurchie:
Updated.

PR #74, opened 2017-05-05
Update getAttribute logic for boolean attributes

mykola-mokhnach:
Please merge it after https://github.com/appium/appium-uiautomator2-server/pull/56 is merged and published

### DIFF
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -1,21 +1,26 @@
 import log from '../logger';
+import _ from 'lodash';
 
 let commands = {}, helpers = {}, extensions = {};
+
+function toBool(s) {
+  return _.isString(s) ? (s.toLowerCase() === 'true') : !!s;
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
@jlipps Exactly. This is the way it is defined in the spec

jlipps:
I am still confused. The end state for a client using the driver looks same as it was before this set of changes. What is the point?

mykola-mokhnach:
@jlipps Read https://github.com/appium/appium/issues/8321

jlipps:
OK I get it now. I had read that, and thought that this change was turning strings into bool for every endpoint. But what I now see this is doing is keeping things bool that were bool and were supposed to be bool, and now letting the string through for the getElementAttribute call. I was confused because now the real change here is in a part of the code which is _not_ in the diff :-)

PR #76, opened 2017-06-22
Add allowInvisibleElements

rjaros87:
Fix for https://github.com/appium/appium-uiautomator2-server/issues/61

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -107,7 +107,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
     this.jwpProxyAvoid = NO_PROXY;
     this.apkStrings = {}; // map of language -> strings obj
 
-    this.settings = new DeviceSettings({ignoreUnimportantViews: false},
+    this.settings = new DeviceSettings({ignoreUnimportantViews: false, allowInvisibleElements: false},
         this.onSettingsUpdate.bind(this));
     // handle webview mechanics from AndroidDriver
     this.chromedriver = null;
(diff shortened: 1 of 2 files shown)


rjaros87:
Correlated with https://github.com/appium/appium-uiautomator2-server/pull/63

lib/driver.js
```diff
@@ -413,14 +413,9 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
-  async setCompressedLayoutHierarchy (compress) {
-    await this.uiautomator2.jwproxy.command('/appium/device/compressedLayoutHierarchy', 'POST', {compressLayout: compress});
-  }
-
   async onSettingsUpdate (key, value) {
-    if (key === "ignoreUnimportantViews") {
-      await this.setCompressedLayoutHierarchy(value);
-    }
+    let settings = {[key]: value};
+    await this.uiautomator2.jwproxy.command('/appium/settings', 'POST', {settings: settings});
```
imurchie: `{settings: settings}` can just be `{settings}` (see https://github.com/getify/You-Dont-Know-JS/blob/master/es6%20%26%20beyond/ch2.md#concise-properties)

sravanmedarapu:
It will be great if you can add test to verify this :)

rjaros87:
@sravanmedarapu test already exists in https://github.com/appium/appium-uiautomator2-driver/blob/master/test/functional/commands/find/by-xpath-e2e-specs.js#L67 for `updateSettings` .
I've added a separate test for `allowInvisibleElemets` starting the Settings application because in the `ApiDemos` app I can't find for example the `statusBarBackground`. I'm not sure, but this might be because of a customized theme in the `ApiDemos`.

PR #77, opened 2017-06-26
Travis fixes

dpgraham:
* Run functional tests on Travis
* Fixed broken tests by waiting for elements to exist before operating on them

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,44 @@
-language: node_js
-node_js:
-  - "7"
-  - "6"
-  - "4"
+language: android
+android:
+  components:
+    # use the latest revision of Android SDK Tools
+    - platform-tools
+    - tools
+
+    # Additional components
+    - extra-google-google_play_services
+    - extra-google-m2repository
+    - extra-android-m2repository
+    - addon-google_apis-google-19
+
+    # Specify at least one system image,
+    # if you need to run emulator(s) during your tests
+    - sys-img-armeabi-v7a-android-22
+    - sys-img-armeabi-v7a-android-17
+env:
+  - _FORCE_LOGS=1 TEST=unit
+  - _FORCE_LOGS=1 TEST=unit/commands
+  - _FORCE_LOGS=1 TEST=functional
+  - _FORCE_LOGS=1 TEST=functional/commands
+  - _FORCE_LOGS=1 TEST=functional/commands/find
+  - _FORCE_LOGS=1 TEST=functional/commands/touch
+before_script:
+  # node stuff
+  - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
+  - nvm install 4
+  - node --version
+  - npm --version
+  - npm install -g gulp
+  - npm install -g mocha
+  - SERVER_INSTALL_ATTEMPTS=50 npm install
+
+  # android stuff
+  - android list targets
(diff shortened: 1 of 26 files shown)


test/functional/commands/attribute-e2e-specs.js
```diff
@@ -15,6 +15,23 @@ let caps = {
   platformName: 'Android'
 };
 
+async function waitForElement (strategy, selector) {
```
jlipps: this could all be replaced with `asyncbox`'s `retryInterval`

```js
async function waitForElement (strategy, selector) {
  return retryInterval(10, 3000, async () => {
    await driver.findElOrEls(strategy, selector);
  };
}
```

or set an implicit wait on the driver first.

.travis.yml
```diff
@@ -1,10 +1,44 @@
...
+
+    # Additional components
+    - extra-google-google_play_services
+    - extra-google-m2repository
+    - extra-android-m2repository
+    - addon-google_apis-google-19
+
+    # Specify at least one system image,
+    # if you need to run emulator(s) during your tests
+    - sys-img-armeabi-v7a-android-22
```
sravanmedarapu: In my opinion `x86`(sys-img-x86-android-22) system images are faster than `armeabi`, but not really sure if travis have the support.

imurchie: Travis doesn't support them, though they are much better. Alas.

test/functional/commands/attribute-e2e-specs.js
imurchie: Or
```js
async function waitForElement (strategy, selector) {
  return await retryInterval(10, 3000, findElOrEls.bind(driver), strategy, selector);
}
```

.travis.yml
```diff
@@ -1,10 +1,44 @@
...
+  - npm install -g gulp
+  - npm install -g mocha
+  - SERVER_INSTALL_ATTEMPTS=50 npm install
+
+  # android stuff
+  - android list targets
+  - echo no | android create avd --force -n test -t android-21 --abi armeabi-v7a
+  - emulator -avd test -no-audio -no-window &
+  - android-wait-for-emulator
+  - adb shell input keyevent 82 &
```
sravanmedarapu: It will be great if we can add `logcat`, it will help us greatly for debugging.

test/functional/commands/attribute-e2e-specs.js
jlipps: good eye Isaac!

amedvedjev:
looks suspicious...
lib/driver.js
line 150 
```
this.opts.appActivity = activity;
this.opts.appActivity = activity;
```

lib/driver.js
```diff
@@ -143,6 +142,16 @@ class AndroidUiautomator2Driver extends BaseDriver {
       };
       _.defaults(this.opts, defaultOpts);
 
+      if (this.isChromeSession) {
+        logger.info("We're going to run a Chrome-based session");
+        let {pkg, activity} = helpers.getChromePkg(this.opts.browserName);
+        this.opts.appPackage = pkg;
+        this.caps.appPackage = pkg;
+        this.opts.appActivity = activity;
+        this.opts.appActivity = activity;
```
jlipps: yeah looks like this should be `this.caps`

imurchie: Yes. I didn't know this was going to be merged right away. Anyway, I have a few more fixes to try out, so I'll fix this there.

PR #80, opened 2017-07-07
Clean up startup and tests

imurchie:
Much of this is just getting the tests running correctly on Travis.

Substantive changes:
1. Adds `uiautomator2ServerInstallTimeout` capability, to configure the amount of time to wait for the instrumentation package to be available on the device
2. Check that the instrumentation package is on the device. On local systems this happened almost immediately. On slower ones (like Travis) it took a while, and so we were already trying to launch it and the launching failed with no instrumentation package available.


lib/uiautomator2.js
```diff
@@ -26,36 +28,80 @@ class UiAutomator2Server {
...
+
+    logger.debug(`Waiting up to ${retries * 1000}ms for instrumentation '${INSTRUMENTATION_TARGET}' to be available`);
+    let output;
+    try {
+      await retryInterval(retries, 1000, async () => {
+        try {
+          output = await this.adb.shell(['pm', 'list', 'instrumentation']);
+          if (output.indexOf('Could not access the Package Manager') !== -1) {
+            output = null; // remove output, so it is not printed below
+            throw new Error(`Problem running package manager: ${output}`);
```
jlipps: the message will have to be constructed before output is set to `null`, no?

lib/uiautomator2.js
```diff
@@ -26,36 +28,80 @@ class UiAutomator2Server {
...
+          } else if (output.indexOf(INSTRUMENTATION_TARGET) === -1) {
+            throw new Error('No instrumentation process found. Retrying...');
+          }
+          logger.debug(`Instrumentation '${INSTRUMENTATION_TARGET}' available`);
+        } catch (err) {
```
jlipps: the fact that we do nothing with `err` other than throw it makes me think we don't need this to be in a try/catch block.

lib/uiautomator2.js
imurchie: Yup.

lib/uiautomator2.js
imurchie: Ah, yes. This was a holdover from the long process of debugging this stuff.

PR #82, opened 2017-07-20
Add test of typing into field with maxLength set

imurchie:
Establish that https://github.com/appium/appium/issues/8279 is no longer a problem.


imurchie:
Seems that maybe it _does_ still affect emulators, at least at lower API levels. I'll do more checking in the morning.

PR #83, opened 2017-07-21
Moved appium-gulp-plugins from devDependencies to dependencies section

sravanmedarapu:
Address install [issue](https://github.com/appium/appium-uiautomator2-server/pull/66#issuecomment-317017483)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "appium-base-driver": "^2.0.18",
     "appium-support": "^2.5.0",
     "appium-uiautomator2-server": "0.x",
+    "appium-gulp-plugins": "^1.3.12",
     "asyncbox": "^2.3.1",
     "babel-runtime": "=5.8.24",
     "bluebird": "^2.10.2",
@@ -59,7 +60,6 @@
   ],
   "devDependencies": {
     "android-apidemos": "^2.1.0",
-    "appium-gulp-plugins": "^1.3.12",
     "appium-test-support": "0.2.0",
     "babel-eslint": "^6.1.2",
     "chai": "^3.4.0",


imurchie:
Is there a better way to do install? For some reason it stopped downloading the server, and obviously my "solution" is a fail.

But I'd rather not have `appium-gulp-plugins` in the dependencies, if possible.

sravanmedarapu:
@imurchie, I could see `Error: Cannot find module 'appium-gulp-plugins'`  while installing driver and suspect because of the [install](https://github.com/appium/appium-uiautomator2-driver/blame/master/package.json#L49)(we are also executing prepublish now as part of install) and when you install the module independently, not be able to identify [appium-gulp-plugins](https://github.com/appium/appium-uiautomator2-driver/blob/36fbc9e487d7f840a52bb99c96d4d7d2c5098b7d/gulpfile.js#L4), Please correct me if my understanding is wrong.

````
> appium-uiautomator2-driver@0.4.4 install /usr/local/lib/node_modules/appium-uiautomator2-driver
> gulp prepublish && node ./bin/install.js

module.js:472
    throw err;
    ^

Error: Cannot find module 'appium-gulp-plugins'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/usr/local/lib/node_modules/appium-uiautomator2-driver/gulpfile.js:4:19)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
````


And regarding downloading the server APKs issue you pointed may be it is because of the network related issue, but I am not able to reproduce.

imurchie:
Yes, that is correct. I'm just wondering if there is a better way to do the install, that would not need to run the prepublish script. I don't know why it stopped downloading the server.

sravanmedarapu:
We don't need to run prepublish script, seems it is recently added to [install](https://github.com/appium/appium-uiautomator2-driver/blob/master/package.json#L49) command. 

Will remove the prepublish then.

imurchie:
Yes, it was added because the server stopped being downloaded, and the install script was not finding the transcribed code.

I am hoping you have a better solution. I have no idea why it stopped working.

sravanmedarapu:
Sorry I got it wrong. May be need to investigate for alternatives. but as a hot fix, If you are OK with we could release with current PR to unblock users to use the latest driver version and will think about the better approach.

imurchie:
yeah, as a quick fix. Can you make sure that running `npm install --production` works?

wangtao169447:
@sravanmedarapu @imurchie 
hi，What is the conclusion？ or the plan is ?

thanks

PR #88, opened 2017-08-02
Refactor server presence verification

mykola-mokhnach:
Make the output cleaner for issues like https://github.com/appium/appium/issues/8956

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -27,20 +27,26 @@ class UiAutomator2Server {
     });
   }
 
+  /**
+   * Installs the apks on to the device or emulator.
+   *
+   * @param {number} installRetries - The count of installation retries
+   */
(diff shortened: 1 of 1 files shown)


lib/uiautomator2.js
```diff
@@ -102,40 +104,18 @@ class UiAutomator2Server {
   }
 
   async getPackageName (apk) {
```
imurchie: Naming here seems to be mixed up. `getPackageName` and `getAPKVersion` take an apk, while `getInstalledPackageVersion` takes a package. Can these be normalized? Perhaps `getAPKName`? The call in `installServerApk` seems to be the only place it is used.

mykola-mokhnach: should be better now

imurchie:
Published in `appium-uiautomator2-driver@0.5.2`

PR #89, opened 2017-08-03
Update eslint

imurchie:


bin/install.js
```diff
@@ -1,22 +1,22 @@
...
+const INTERVAL = 1500;
+let attemptedToBuild = false;
 
 function doInstall () {
   // UiAutomator2 needs Java. Fail early if it doesn't exist
-  var androidHelpers = require('appium-android-driver').androidHelpers;
-  androidHelpers.getJavaVersion().then(function () {
-    var tries = 0;
+  let androidHelpers = require('appium-android-driver').androidHelpers;
+  androidHelpers.getJavaVersion().then(function () { // eslint-disable-line promise/prefer-await-to-then
```
jlipps: should we make this exception file-wide since we can't do promises in this whole file?

lib/uiautomator2.js
```diff
@@ -143,14 +143,14 @@ class UiAutomator2Server {
   async startSessionUsingAdbKit (deviceUDID) {
     logger.info(`Running command: 'adb -s ${deviceUDID} shell am instrument -w ${INSTRUMENTATION_TARGET}'`);
     this.client.shell(deviceUDID, `am instrument -w ${INSTRUMENTATION_TARGET}`)
-      .then(adbkit.util.readAll)
-      .then(function (output) {
+      .then(adbkit.util.readAll) // eslint-disable-line promise/prefer-await-to-then
```
jlipps: why don't we just convert this into a series of `await` calls?

imurchie: No idea!

imurchie: Ah. Turns out I did have an idea and forgot. 

Anyway, we're not actually waiting here at all. Just setting up the logging from the adb command, which runs in the background the whole time. We _could_ separate out the `adbkit.util.readAll` command but it would remove the error logging.

PR #91, opened 2017-08-16
Save default IME so we can restore

imurchie:
Save the IME that is in use before switching to Appium's Unicode keyboard. Analogous to https://github.com/appium/appium-android-driver/blob/master/lib/driver.js#L224

Fixes https://github.com/appium/appium/issues/8755


imurchie:
Published as `appium-uiautomator2-driver@0.5.4`.

PR #92, opened 2017-08-29
Use chromedriver for Chrome sessions

dpgraham:
* Call 'startChromeSession' on chrome sessions
* Refactor URL tests to use 'wd' instead of internal commands

(connected to: https://github.com/appium/appium-uiautomator2-driver/pull/92)

### DIFF
--- a/test/functional/commands/url-e2e-specs.js
+++ b/test/functional/commands/url-e2e-specs.js
@@ -1,8 +1,8 @@
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { BROWSER_CAPS } from '../desired';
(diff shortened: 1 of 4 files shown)


test/functional/commands/url-e2e-specs.js
```diff
@@ -1,8 +1,8 @@
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { BROWSER_CAPS } from '../desired';
-import ADB from 'appium-adb';
-import { initDriver } from '../helpers/session';
+import {  startServer, DEFAULT_PORT } from '../../..';
```
jlipps: one too many spaces here

PR #100, opened 2017-10-04
Add `/log/types` endpoint to NO_PROXY

vmaxim:
`/log/types` endpoint implemented in `appium-android-driver` and should not  be proxied to `appium-uiautomator2-server`.


imurchie:
Published in `appium-uiautomator2-driver@0.6.6`.

PR #106, opened 2017-10-31
Throw an exception if running on an unsupported OS

mykola-mokhnach:
I saw several issue reports about UIA2 is not working properly and then it turns out that people are trying to run it on Android 4.4 or even older versions. This PR should improve the situation.


imurchie:
Published in `appium-uiautomator2-driver@0.7.2`.

PR #101, opened 2017-10-04
UiScrollable tests should perform fair scrolling - not a single swipe.

vmaxim:
Note: with this changes all UiScrollable tests will fail - as it should be, due to issue appium/appium#9328.

PR #107, opened 2017-11-01
Adding capabilities for visual testing of native apps

danielputerman:
This PR is part of a number of PRs for Appium (additional ones are for appium-base-driver, uiautomator2-server, and ios-driver), adding capabilities which allow for visual testing of native apps. These capabilities include:

Viewport related capabilities: status bar height, viewport rect (position/size), getting viewport screenshot.
Additional functionality for scrollable elements: finding scrollable views in the application, retrieving their content size and providing deterministic scrolling inside them.

Specifically this PR includes the following functionality:
Getting the viewport screenshot (no status-bar, no bottom buttons).
Retrieving additional viewport properties as part of the session details (status bar height, viewport rect and the device pixel ratio).
Functionality relating to scrollable elements: finding the first visible item, and the content size of a scrollable element.


lib/commands/find.js
```diff
@@ -6,6 +6,10 @@ let helpers = {}, extensions = {};
  * this.element initialized in find.js of appium-android-drive.
  */
 helpers.doFindElementOrEls = async function (params) {
+  if (params.strategy === "xpath" && params.selector === ("//*[@onScreen=true]")) {
```
mykola-mokhnach: this code block looks very hacky to me...

lib/commands/general.js
```diff
@@ -47,6 +47,22 @@ commands.getWindowSize = async function () {
...
 
+commands.execute = async function (script, args) {
+  if (script.match(/^mobile\:/)) {
+    script = script.replace(/^mobile\:/, '').trim();
+    if (script === 'scroll' && !this.isWebContext()) {
+      let elementId = args[0].element;
+      return await this.uiautomator2.jwproxy.command(`/appium/element/${elementId}/scroll_to`, 'POST', {});
+    } else if (script === 'viewportScreenshot' && !this.isWebContext()) {
+      return await this.getViewportScreenshot();
+    } else {
```
mykola-mokhnach: these two `else` blocks are not needed. this.executeCommand after `if` block is enough, since the internal condition anyway perform _return_

lib/commands/viewport.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+commands.getViewPortRect = async function () {
+  const windowSize = await this.getWindowSize();
+  const statusBarHeight = await this.getStatusBarHeight();
+  return {left: 0, top: statusBarHeight, width: windowSize.width, height: windowSize.height - statusBarHeight};
+};
+
+commands.getFirstVisible = async function (elementId) {
+  return await this.uiautomator2.jwproxy.command(`/appium/element/${elementId}/first_visible`, 'GET', {});
```
mykola-mokhnach: What the purpose of this endpoint?

test/functional/commands/viewport/viewport-e2e-specs.js
```diff
@@ -0,0 +1,70 @@
...
+    scrollableElementId = element.ELEMENT;
+    scrollableElementId.should.not.equal(null);
+  });
+
+  it('should get content size from scrollable element', async () => {
+    let contentSize = await driver.getAttribute("contentSize", scrollableElementId);
+    contentSize.should.not.equal(null);
+  });
+
+  let firstElementId;
```
vmaxim: Please remove dependency between tests.

test/functional/commands/viewport/viewport-e2e-specs.js
```diff
@@ -0,0 +1,70 @@
...
+  it('should get first element from scrollable element', async () => {
+    let element = await driver.findElOrEls('xpath', '//*[@onScreen=true]', false, scrollableElementId);
+    firstElementId = element.ELEMENT;
+    firstElementId.should.not.equal(null);
+  });
+
+  it('should scroll to element by provided elementId', async () => {
+    let result = await driver.execute('mobile: scrollTo', [{element: firstElementId}]);
```
vmaxim: I think it's a bad idea to test scrolling against first element.

lib/driver.js
```diff
@@ -201,6 +203,17 @@ class AndroidUiautomator2Driver extends BaseDriver {
     }
   }
 
+  async getSession () {
+    let caps = await super.getSession();
+    const pixelRatio = await this.getDevicePixelRatio();
+    const statBarHeight = await this.getStatusBarHeight();
+    const viewportRect = await this.getViewPortRect();
```
vmaxim: I think this data can be cached on start session.

lib/commands/viewport.js
```diff
@@ -0,0 +1,33 @@
+
+import { ImageHelpers } from 'appium-base-driver';
+
+let extensions = {}, commands = {};
+
+commands.getStatusBarHeight = async function () {
+  return await this.uiautomator2.jwproxy.command(`/appium/device/stat_bar_height`, 'GET', {});
```
vmaxim: [/wd/hub/session/:sessionId/appium/device/system_bars](https://github.com/appium/appium-base-driver/blob/8481b9ac7cfa89563e86bc9f53e1b1d0127f6d6b/lib/mjsonwp/routes.js#L418)

lib/commands/general.js
```diff
@@ -47,6 +47,19 @@ commands.getWindowSize = async function () {
...
+  if (script.match(/^mobile\:/)) {
+    script = script.replace(/^mobile\:/, '').trim();
+    if (script === 'scroll' && !this.isWebContext()) {
+      let elementId = args[0].element;
+      return await this.uiautomator2.jwproxy.command(`/appium/element/${elementId}/scroll_to`, 'POST', {});
+    } else if (script === 'viewportScreenshot' && !this.isWebContext()) {
+      return await this.getViewportScreenshot();
+    }
+  }
+  this.executeCommand(script, args);
```
imurchie: I don't think this is what should be done at this point. This is a completely new function for this driver, so there is no other behaviour that should be accounted for. 

`executeCommand` is the central command executor for the system, putting commands into the queue and handling everything about command execution. Calling it here will almost certainly break things in ways that would be difficult to fix.

See https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/driver.js#L195-L270

imurchie: We do need to handle the web context, though.

lib/commands/general.js
```diff
@@ -47,6 +47,19 @@ commands.getWindowSize = async function () {
   return await this.uiautomator2.jwproxy.command('/window/current/size', 'GET', {});
 };
 
+commands.execute = async function (script, args) {
+  if (script.match(/^mobile\:/)) {
+    script = script.replace(/^mobile\:/, '').trim();
+    if (script === 'scroll' && !this.isWebContext()) {
```
imurchie: These conditionals seem weird. We can dispatch with the web context in one place and then just work on the value of `script`:
```js
if (this.isWebContext()) {
  // proxy to chromedriver
  return;
}
if (script.match(/^mobile\:/)) {
  let mobileScript = script.replace(/^mobile\:/, '').trim();
  if (mobileScript === 'scroll') { // why 'scroll' and not 'scrollTo'?
    let elementId = args[0].element;
    return await this.uiautomator2.jwproxy.command(`/appium/element/${elementId}/scroll_to`, 'POST', {});
  } else if (mobileScript === 'viewportScreenshot') {
    return await this.getViewportScreenshot();
  }
}

throw new Error(`Unable to execute unknown script: '${script}'`);
```

test/functional/commands/viewport/viewport-e2e-specs.js
```diff
@@ -62,9 +62,4 @@ describe('testViewportCommands', function () {
     firstElementId = element.ELEMENT;
     firstElementId.should.not.equal(null);
   });
-
-  it('should scroll to element by provided elementId', async () => {
```
vmaxim: Why you delete this test? You don't want to test new endoiint? :)

lib/commands/viewport.js
danielputerman: I wasn't aware of this link. Notice however, that this route is not implemented for appium-ios-driver, while ours is implemented for both Android and iOS. Should we still use this route?

danielputerman: @vmaxim @imurchie  ?

jlipps: somehow I wasn't aware of this route either. But given that it exists we should probably use it instead of creating another route. @danielputerman it looks possible to simply update the return value to conform to the shape of `system_bars`, right?

lib/driver.js
```diff
@@ -94,6 +94,8 @@ const NO_PROXY = [
   ['GET', new RegExp('^/session/[^/]+/ime/activated')],
   ['POST', new RegExp('^/session/[^/]+/ime/deactivate')],
   ['POST', new RegExp('^/session/[^/]+/ime/activate')],
+  ['GET', new RegExp('^/session/[^/]+/appium/viewport_screenshot')],
```
jlipps: didn't see this route in the basedriver PR; is it forthcoming?

lib/driver.js
```diff
@@ -196,13 +198,22 @@ class AndroidUiautomator2Driver extends BaseDriver {
       this.opts.adbPort = this.opts.adbPort || DEFAULT_ADB_PORT;
 
       await this.startUiAutomator2Session();
+      if (!caps.unitTests) {
```
jlipps: it would be better not to have test conditionals in production code. instead, we could simply stub out `fillDeviceDetails` to do nothing during unit tests

test/unit/driver-specs.js
```diff
@@ -21,7 +21,7 @@ describe('driver.js', () => {
   describe('createSession', () => {
     it('should throw an error if app can not be found', async () => {
       let driver = new AndroidUiautomator2Driver({}, false);
```
jlipps: yes, better to stub out `driver.fillDeviceDetails` here rather than add a cap used just for testing

jlipps:
@imurchie @mykola-mokhnach this one is also ready for final review (and depends on https://github.com/appium/appium-uiautomator2-server/pull/86), which would need to be published and then `installer.js` here also updated to rely on the new server.

lib/commands/general.js
```diff
@@ -47,6 +47,21 @@ commands.getWindowSize = async function () {
   return await this.uiautomator2.jwproxy.command('/window/current/size', 'GET', {});
 };
 
+commands.execute = async function (script, args) {
```
mykola-mokhnach: this might override the magic "mobile: shell" endpoint. can we somehow merge the stuff?

lib/commands/viewport.js
```diff
@@ -0,0 +1,36 @@
...
+
+let extensions = {}, commands = {};
+
+commands.getStatusBarHeight = async function () {
+  const {statusBar} = await this.uiautomator2.jwproxy.command(`/appium/device/system_bars`, 'GET', {});
+  return statusBar;
+};
+
+// this function is memoized in the constructor
+commands.getDevicePixelRatio = async function () {
```
mykola-mokhnach: this call is actually not used, right?

lib/commands/general.js
jlipps: hmmm, that one was added while working on this code. i'll check it out and make sure they don't clash.

lib/commands/viewport.js
jlipps: it is used during session initialization to populate the session capabilities

jlipps:
ok @imurchie @mykola-mokhnach the server side has been incorporated into this PR so hopefully we'll get some CI feedback, and the code is ready for final review

test/functional/commands/touch-action-e2e-specs.js
```diff
@@ -77,19 +78,24 @@ describe('apidemo - touch', function () {
 
   describe('mobile: scrollBackTo', function () {
     let driver;
-    before(async () => {
+    before(async function () {
+      if (await isArmEmu()) {
+        // arm emulators are too slow and android's scrollToBeginning is broken
+        // on them for some reason
```
vmaxim: Are you sure this is an ARM problem? Did you tried to digging into this issue?

jlipps: i did try digging in. the issue is that android thinks it is at the top of the view and returns as such, even when this is not the case. i could only reproduce it on arm emus. i don't know if it has anything to do with arm itself. probably not. i wonder if it has to do with the slowness causing parts of the gesture to not be recognized, or cause some other confusion.

in short, i don't fully understand the root cause, but given that it only seems to affect a class of emus no one actually uses, except us in CI, i felt fine skipping.

jlipps:
@imurchie the keyboard tests failed here, despite that they passed a few commits ago and the only change was merging from master. is master working?

jlipps:
going to restart that travis job and see if it was some kind of flake

jlipps:
@imurchie ok, appears to be a flake unrelated to this change. mind giving a final pass and the green checkmark when you have a moment?

test/functional/commands/touch-action-e2e-specs.js
vmaxim: This is strange that you can reproduce this issue on arm emu only, because all scrolling stuff are currently broken on uia2 - appium/appium#9328, #101. Can you repro this issue without ```NotificationListener```?

jlipps: interesting, @vmaxim. i didn't realize uiauto2 could steal notifications from other listeners. that may indeed be what's happening. either way it seems unrelated to this PR, for which quarantining the test for environments it's broken in is reasonable, as long as we don't forget to fix the underlying problem and unquarantine this test when that is resolved.

jlipps:
published this in `1.0.0` since we should be out of beta with this driver.

imurchie:
\o/

PR #108, opened 2017-11-02
Add systemPort to the list of desired caps

mykola-mokhnach:
This is needed to prevent "unknown capability" error from base driver.


imurchie:
Published in `appium-uiautomator2-driver@0.7.3`.

PR #109, opened 2017-11-03
`appium-uiautomator2-driver` should use own `clear` endpoint,...

vmaxim:
…not a tricky `clear` command from `appium-android-driver`


imurchie:
Is there a reason for this? The Android driver clear function is the product of a _lot_ of problems over the years.

vmaxim:
I can't see that we gave a chance to execute `UiObject2.clear` command  [before](https://github.com/appium/appium-android-driver/pull/156/files#diff-dc10b0f0de22327416bd1e9ef8c6b8e9L68).
The implementation of `clear` command in [`uia2`](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject2.java#337) is much differ than [`uia`](https://android.googlesource.com/platform/frameworks/uiautomator/+/android-support-test/src/main/java/android/support/test/uiautomator/UiObject.java#684).
If you worry about regression I can test it against our production testing - it's about more that 300 test cases with runtime over than 12 hours.

jlipps:
@vmaxim that would be incredible!

vmaxim:
Found regression with a password text field. Closing this PR until the issue is fixed.
Also e2e tests should be updated appium/appium-uiautomator2-driver/pull/111

vmaxim:
Now it should be ok

PR #112, opened 2017-11-17
TestObject fixes

dpgraham:
* Fixed some tests that were broken in TestObject (some T.O. tests are still broken but I think we should make that a separate issue)
* Use AppVeyor matrix of platform versions instead of hardcoding by device (Android 6, 7 and 7.1)


test/functional/commands/keyboard/keyboard-e2e-specs.js
```diff
@@ -35,7 +35,7 @@ function deSamsungify (text) {
 }
 
 async function getElement (driver, className) {
-  return await retryInterval(10, 1000, async () => {
+  return await retryInterval(process.env.TESTOBJECT_E2E_TESTS ? 100 : 10, 1000, async () => {
```
mykola-mokhnach: wow, isn't it a bit too much?:)

dpgraham: Perhaps, but it did seem to fix it.

PR #116, opened 2017-12-11
Add deviceScreenDensity to session caps

fuzzybinary:
This also updates to lodash 4, as I was having issues with conflicts with other appium packages.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -260,11 +260,12 @@ class AndroidUiautomator2Driver extends BaseDriver {
     // and get it onto our 'opts' object so we use it from now on
     Object.assign(this.opts, appInfo);
 
-    // set actual device name, udid, platform version, screen size, model and manufacturer details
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -265,6 +265,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
     this.caps.deviceUDID = this.opts.udid;
     this.caps.platformVersion = await this.adb.getPlatformVersion();
     this.caps.deviceScreenSize = await this.adb.getScreenSize();
+    this.caps.deviceScreenDensity = await this.adb.getScreenDensity();
```
mykola-mokhnach: perhaps, the comment above should be updated

fuzzybinary: Fixed and force pushed for a clean history ;)

imurchie:
Published in `appium-uiautomator2-driver@0.8.0`

PR #121, opened 2017-12-21
getProxyAvoidList should not overwrite additional rule

neofreko:
**Issue:**
nativeWebScreenshot preference was overwritten on each getProxyAvoidList call. This makes screenshot command getting proxied to chromedriver when user is on webview mode. Which most likely will timeout.

**Implementation:**
Fix test cases to actually check proxyAvoidList content by feeding in `/session/zzz/screenshoot/`.
Former test for nativeWebScreenshot had false positive since NO_PROXY list is already over 40 items.
Add additional test cases to specify native mode behavior which ignores nativeWebScreenshot setting.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -134,7 +134,6 @@ class AndroidUiautomator2Driver extends BaseDriver {
     this.jwpProxyActive = false;
     this.defaultIME = null;
     this.jwpProxyAvoid = NO_PROXY;
-    this.additionalJwpProxyAvoid = [];
     this.apkStrings = {}; // map of language -> strings obj
 
     this.settings = new DeviceSettings({ignoreUnimportantViews: false, allowInvisibleElements: false},
@@ -178,10 +177,6 @@ class AndroidUiautomator2Driver extends BaseDriver {
         this.opts.appPackage = this.caps.appPackage = pkg;
         this.opts.appActivity = this.caps.appActivity = activity;
         logger.info(`Chrome-type package and activity are ${pkg} and ${activity}`);
-      }
-
-      if (this.opts.nativeWebScreenshot) {
-        this.additionalJwpProxyAvoid.push(['GET', new RegExp('^/session/[^/]+/screenshot')]);
       }
 
       if (this.opts.reboot) {
@@ -561,7 +556,9 @@ class AndroidUiautomator2Driver extends BaseDriver {
     } else {
       this.jwpProxyAvoid = NO_PROXY;
     }
-    this.jwpProxyAvoid = this.jwpProxyAvoid.concat(this.additionalJwpProxyAvoid);
+    if (this.opts.nativeWebScreenshot) {
+      this.jwpProxyAvoid = [...this.jwpProxyAvoid, ['GET', new RegExp('^/session/[^/]+/screenshot')]];
+    }
 
     return this.jwpProxyAvoid;
   }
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -560,6 +561,8 @@ class AndroidUiautomator2Driver extends BaseDriver {
     } else {
       this.jwpProxyAvoid = NO_PROXY;
     }
+    this.jwpProxyAvoid = this.jwpProxyAvoid.concat(this.additionalJwpProxyAvoid);
```
mykola-mokhnach: I think we don't need `this.additionalJwpProxyAvoid` variable. Just change this line to

```js
if (this.opts.nativeWebScreenshot) {
   this.jwpProxyAvoid = [...this.jwpProxyAvoid, ['GET', new RegExp('^/session/[^/]+/screenshot')]];
}
```

test/unit/driver-specs.js
```diff
@@ -119,7 +119,8 @@ describe('driver.js', () => {
         }).should.throw;
       });
       describe('nativeWebScreenshot', function () {
-        let listLength;
+        let list;
```
mykola-mokhnach: can we have a better name for this variable?

test/unit/driver-specs.js
```diff
@@ -119,7 +119,8 @@ describe('driver.js', () => {
...
+        let nativeWebScreenshotFilter = item => { return item[0] === "GET" && item[1].test('/session/xxx/screenshot/');};
```
mykola-mokhnach: change to 
`item => item[0] === 'GET' && item[1].test('/session/xxx/screenshot')`

test/unit/driver-specs.js
```diff
@@ -130,14 +131,39 @@ describe('driver.js', () => {
...
+
+        describe('on native mode', function () {
+          beforeEach(function () {
+            driver.chromedriver = null;
+          });
+          it('should never proxy screenshot regardless of nativeWebScreenshot setting (on)', async function () {
+            // nativeWebScreenshot on
+            await driver.createSession({platformName: 'Android', deviceName: 'device', browserName: 'chrome', nativeWebScreenshot: true});
+            list = driver.getProxyAvoidList().filter(nativeWebScreenshotFilter);
+            list.should.not.empty;
```
mykola-mokhnach: -> not.be.empty

test/unit/driver-specs.js
```diff
@@ -130,14 +131,39 @@ describe('driver.js', () => {
...
+            list.should.not.be.empty;
+          });
         });
-        it('should not proxy screenshot if nativeWebScreenshot is on', async function () {
-          await driver.createSession({platformName: 'Android', deviceName: 'device', browserName: 'chrome', nativeWebScreenshot: true});
-          driver.getProxyAvoidList().should.have.length(listLength + 1);
+
+        describe('on native mode', function () {
+          beforeEach(function () {
+            driver.chromedriver = null;
```
mykola-mokhnach: should we restore the original value of `driver.chromedriver` after each test?

lib/driver.js
neofreko: Oh, I didn't realize `this.opts` is still available :D. Let me rectify.

test/unit/driver-specs.js
neofreko: I actually confused about how current tests share `driver` variable. We may want to make sure each test has a clean `driver`.

imurchie: These particular tests only share the variable. Before each test execution there is a new driver object instantiated (see line 125). There is no need to restore anything. There is also no need to set `driver.chromedriver` to anything, since it will already be undefined.

imurchie:
Published in `appium-uiautomator2-driver@0.10.0`

PR #129, opened 2018-01-17
fix JWProxy opt key. Also pass adb host to adbkit

rajdeepv:
This change is needed if we are running tests on a device connected to the remote machine. 
1)JWProxy uses  :server instead of :host
2) tell adbkit about remoteAdbHost instead of assuming localhost. This fixes starting instrumentation on remote machine with 'adb shell am instrument..'


imurchie:
Published in `appium-uiautomator2-driver@1.0.1`

PR #135, opened 2018-01-22
Remove port forward before killing emulator

vmaxim:


lib/driver.js
```diff
@@ -485,6 +485,15 @@ class AndroidUiautomator2Driver extends BaseDriver {
         }
       }
       await this.adb.stopLogcat();
+      if (this.opts.systemPort !== undefined) {
```
mykola-mokhnach: I'd rather use util.hasValue helper

PR #140, opened 2018-02-08
Bump android driver, and fix chrome

imurchie:
Update the version of `appium-android-driver` and use the proper function for chrome initialization.


mykola-mokhnach:
Some functional tests didn't like the change :(

PR #145, opened 2018-02-20
Properly override setValueImmediate API call

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10211


shridharkalagi:
@mykola-mokhnach  Fix doesn't seem to be working. Slider isn't moving at all now with .setValue() function. Find the server logs [here](https://gist.github.com/shridharkalagi/beddaa857f2e8cee1f27c40ac71b3278).

PR #146, opened 2018-02-23
Disable testobject

dpgraham:
Disabling T.O. until we switch over to using production instead of staging

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,46 +1,46 @@
-environment:
-  nodejs_version: "6"
-  GH_TOKEN:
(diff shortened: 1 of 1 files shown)


imurchie:
I presume that this is just commented out, instead of deleted, so that it is there for later when we move to TO production.

Would it be better to just get rid of the AppVeyor webhook, and disable that way?

mykola-mokhnach:
There is still some instability in Travis functional tests:

```
  1) createSession default adb port should get device model, manufacturer and screen size in session details:
     TypeError: Cannot read property 'should' of null
```

dpgraham:
I couldn't figure out how to do that @imurchie; do you know how to do that?

PR #147, opened 2018-02-27
Optimise server apk install for UIA2 server

mykola-mokhnach:
The previous implementation mostly duplicates what is already present in `adb.installOrUpgrade` method

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -5,12 +5,16 @@ import { SERVER_APK_PATH as apkPath, TEST_APK_PATH as testApkPath, version as se
 import adbkit from 'adbkit';
 import { getRetries } from './utils';
 import { util } from 'appium-support';
+import B from 'bluebird';
 
 
 const REQD_PARAMS = ['adb', 'tmpDir', 'host', 'systemPort', 'devicePort', 'disableWindowAnimation'];
 const SERVER_LAUNCH_RETRIES = 20;
 const SERVER_INSTALL_RETRIES = 20;
 const INSTRUMENTATION_TARGET = 'io.appium.uiautomator2.server.test/android.support.test.runner.AndroidJUnitRunner';
+const SERVER_PACKAGE_ID = 'io.appium.uiautomator2.server';
+const SERVER_TEST_PACKAGE_ID = `${SERVER_PACKAGE_ID}.test`;
+
 
 class UiAutomator2Server {
   constructor (opts = {}) {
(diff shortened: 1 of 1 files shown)


lib/uiautomator2.js
```diff
@@ -35,33 +39,31 @@ class UiAutomator2Server {
...
-        await this.uninstallServerAndTest(apkPackage, testApkPackage);
-      }
-    }
-
-    if (!isApkInstalled) {
-      await this.signAndInstall(apkPath, apkPackage, installTimeout);
-    }
-    if (!isTestApkInstalled) {
-      await this.signAndInstall(testApkPath, testApkPackage, installTimeout, true);
+    const packagesInfo = [
```
jlipps: just a style thing, but for lists of objects, I prefer:

```js
const foo = [{
  key1: 'blah',
}, {
  key1: 'blarg',
}];
```

lib/uiautomator2.js
```diff
@@ -35,33 +39,31 @@ class UiAutomator2Server {
...
+        appPath: apkPath,
+        appId: SERVER_PACKAGE_ID,
+      },
+      {
+        appPath: testApkPath,
+        appId: SERVER_TEST_PACKAGE_ID,
+      },
+    ];
+    if (await B.reduce(packagesInfo,
+                       async (accumulator, pkgInfo) => (await this.checkAndSignCert(pkgInfo.appPath, pkgInfo.appId)) || accumulator,
```
jlipps: maybe breaking out the accumulator method would make this more readable?

lib/uiautomator2.js
```diff
@@ -35,33 +39,31 @@ class UiAutomator2Server {
...
+      {
+        appPath: apkPath,
+        appId: SERVER_PACKAGE_ID,
+      },
+      {
+        appPath: testApkPath,
+        appId: SERVER_TEST_PACKAGE_ID,
+      },
+    ];
+    if (await B.reduce(packagesInfo,
```
jlipps: so this will be true if one of the apks was resigned?

lib/uiautomator2.js
```diff
@@ -35,33 +39,31 @@ class UiAutomator2Server {
...
+                       async (accumulator, pkgInfo) => (await this.checkAndSignCert(pkgInfo.appPath, pkgInfo.appId)) || accumulator,
+                       false)) {
+      await B.each(packagesInfo, async (pkgInfo) => {
```
jlipps: rather than `B.each`, we typically just use a `for` loop. `B.each` is serial so it provides not advantages over a for loop and is less readable IMO

lib/uiautomator2.js
mykola-mokhnach: accumulator is just a value. The idea is that if any of binaries is unsigned then we need to resign them and uninstall from the device

lib/uiautomator2.js
mykola-mokhnach: ^

lib/uiautomator2.js
mykola-mokhnach: ok

lib/uiautomator2.js
jlipps: makes sense. my only issue now is it looks ugly :-)

PR #148, opened 2018-03-21
Add retry to startApp method call

dpgraham:
* Now that we use `apkAnalyzer` to parse activity names, it's not always going to be the case that an activity name will be "fully qualified"
* We need to set retry to `true` again so that if the activity does not have '.' in front of it, try again with the dot
  * e.g.: If running activity `io.appium.android.apis/ApiDemos` does not work, make it try again with `io.appium.android.apis/.ApiDemos`

(related to https://github.com/appium/appium-adb/pull/308)

(connected to issue https://github.com/appium/appium/issues/10381)

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -397,7 +397,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
         waitActivity: this.opts.appWaitActivity,
         optionalIntentArguments: this.opts.optionalIntentArguments,
         stopApp: !this.opts.dontStopAppOnReset,
-        retry: false
+        retry: true
       });
     }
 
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
It looks like some strings tests are now broken. I'll look into it

dpgraham:
@mykola my other PR will fix those

mykola-mokhnach:
Nah it's my party. I broke it and I will return it back into working state :-P

https://github.com/appium/appium-android-driver/pull/344
https://github.com/appium/appium-uiautomator2-driver/pull/149
https://github.com/appium/appium-adb/pull/309

jlipps:
🕺 🎵 🎂   it's my party and i can fix it if i want to 🎂 🎵 🕺

PR #151, opened 2018-03-23
Add deep link command

dpgraham:
* Use `mobile: deepLink` as the command name
* Requires a URL and Package
* Will only work if instant apps are enabled on the device/emulator, otherwise it throws an exception

(connected to: https://github.com/appium/appium/issues/10329)

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -90,14 +90,7 @@ commands.setUrl = async function (url) {
 
 commands.mobileDeepLink = async function (opts = {}) {
   const {url, package:pkg} = opts;
-
-  const res = await this.adb.startUri(url, pkg);
-
-  // This command never seems to return a non-zero status code, so check for this phrase
-  // to see if it didn't work
-  if (res.toLowerCase().includes('unable to resolve intent')) {
-    throw new Error(res);
-  }
+  return await this.adb.startUri(url, pkg);
 };
 
 commands.openNotifications = async function () {
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -65,6 +65,7 @@ extensions.executeMobile = async function (mobileCommand, opts = {}) {
     shell: async (x) => await this.mobileShell(x),
     scrollBackTo: async (x) => await this.mobileScrollBackTo(x),
     viewportScreenshot: async (x) => await this.mobileViewportScreenshot(x),
+    deepLink: async (x) => await this.mobileDeepLink(x),
   };
 
   if (!_.has(mobileCommandsMapping, mobileCommand)) {
@@ -85,6 +86,18 @@ commands.mobileViewportScreenshot = async function () {
 
 commands.setUrl = async function (url) {
   await this.adb.startUri(url, this.opts.appPackage);
+};
+
+commands.mobileDeepLink = async function (opts = {}) {
+  const {url, package:pkg} = opts;
+
+  const res = await this.adb.startUri(url, pkg);
+
+  // This command never seems to return a non-zero status code, so check for this phrase
+  // to see if it didn't work
+  if (res.toLowerCase().includes('unable to resolve intent')) {
+    throw new Error(res);
+  }
 };
 
 commands.openNotifications = async function () {
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "adbkit": "^2.4.1",
+    "appium-adb": "^6.5.2",
     "appium-android-driver": "^2.1.3",
     "appium-base-driver": "^2.18.4",
     "appium-support": "^2.12.0",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
   },
   "dependencies": {
     "adbkit": "^2.4.1",
-    "appium-adb": "^6.0.1",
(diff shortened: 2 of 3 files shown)


jlipps:
we could also just override the `url` command for deep linking purposes, and assume the package is the currently-running one? or what's the primary use case here?

lib/commands/general.js
```diff
@@ -87,6 +88,18 @@ commands.setUrl = async function (url) {
   await this.adb.startUri(url, this.opts.appPackage);
 };
 
+commands.mobileDeepLink = async function (opts = {}) {
+  const {url, package:pkg} = opts;
```
mykola-mokhnach: what is the point to rename pkg->package?

lib/commands/general.js
```diff
@@ -87,6 +88,18 @@ commands.setUrl = async function (url) {
...
+
+  const res = await this.adb.startUri(url, pkg);
+
+  // This command never seems to return a non-zero status code, so check for this phrase
+  // to see if it didn't work
+  if (res.toLowerCase().includes('unable to resolve intent')) {
```
mykola-mokhnach: should this verification be inside `statrtUri` ?

lib/commands/general.js
dpgraham: Package is a reserved word and breaks the linter

dpgraham:
@jlipps It's for automating "Instant Apps".

* Instants Apps work by clicking a deep link and if the app isn't installed, it automatically opens a miniature version of the app. 
* We wouldn't be able to have the `appPackage` in the capabilities because if the app isn't installed, then the session will fail to start and we want to be able to handle the use case where we click a link for an app that isn't already installed

lib/commands/general.js
dpgraham: The `unable to resolve intent` thing only seems to happen in the case where a deep link is invoked and the package isn't installed and there's no instant app to open.... but I suppose it wouldn't hurt to just include this in the `startUri` method.

dpgraham:
Fixes pushed

test/functional/commands/general/mobile-command-e2e-specs.js
```diff
@@ -6,19 +6,47 @@ import { initDriver } from '../../helpers/session';
...
+
+        // Check that the source has the package name somewhere
+        await driver.source().should.eventually.match(/com\.move\.realtor/);
+
+        // Check that we can find a native element and interact with it
+        const btn = await driver.elementsByXPath('//android.widget.Button');
+        btn.length.should.be.above(0);
+        await btn[0].click();
+      } catch (e) {
+        e.message.should.match(/unable to resolve intent/i);
```
jlipps: i don't think a test should have two possible different verification scenarios. either it should test the success or it should test the failure deterministically. here it's testing one or the other and we'll never know which.

dpgraham: Yeah I had the same thought, I"m just not sure how to go about determining if the Android Emulator that we're testing on supports Instant Apps.

dpgraham:
(btw: the failing test should pass once new `appium-adb` is published)

test/functional/commands/general/mobile-command-e2e-specs.js
dpgraham: @jlipps I just changed it so that _only_ execute is inside the try-catch block. It occurred to me that having assertions inside of a try-catch block is probably bad practice.

package.json
```diff
@@ -27,7 +27,6 @@
   },
   "dependencies": {
     "adbkit": "^2.4.1",
-    "appium-adb": "^6.0.1",
```
imurchie: Why is this being removed? `appium-adb` is still used in this package. Unfortunately linting doesn't complain this because of npm 3's flat dependency structure.

test/functional/commands/general/mobile-command-e2e-specs.js
```diff
@@ -6,19 +6,48 @@ import { initDriver } from '../../helpers/session';
...
+        e.message.should.match(/Original error: Appium server must have relaxed security flag set in order to run any shell commands/);
+      }
+    });
+  });
+  describe('mobile:deepLink', function () {
+    it('should be able to launch apps using Instant Apps', async function () {
+      try {
+        await driver.execute("mobile: deepLink", {url: 'https://www.realtor.com/realestateandhomes-search/San-Jose_CA', package: 'com.move.realtor'});
+      } catch (e) {
+        e.message.should.match(/unable to resolve intent/i);
```
imurchie: Elsewhere we ensure that there is a particular emulator available which is capable of handling the case under test. See, for example, https://github.com/appium/appium-android-driver/blob/master/test/functional/chrome-e2e-specs.js

package.json
dpgraham: That was a mistake.

test/functional/commands/general/mobile-command-e2e-specs.js
dpgraham: I don't think there will be any emulators that can do this, actually. There's a setting that you have to enable on a device.

imurchie: That should be noted, so the test _can_ be run to make sure it works.

PR #152, opened 2018-03-23
Add mobile: handlers for getting and setting clipboard content

mykola-mokhnach:
Depends on https://github.com/appium/appium-uiautomator2-server/pull/124

### DIFF
--- a/lib/commands/clipboard.js
+++ b/lib/commands/clipboard.js
@@ -1,0 +1,21 @@
+import log from '../logger';
+import { util } from 'appium-support';
+
+
+let extensions = {}, commands = {};
+
+commands.mobileSetClipboard = async function (opts = {}) {
+  if (!util.hasValue(opts.content)) {
+    log.errorAndThrow(`The 'content' argument is mandatory`);
+  }
+  return await this.uiautomator2.jwproxy.command('/appium/clipboard/set', 'POST', opts);
+};
+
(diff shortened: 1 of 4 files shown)


lib/commands/clipboard.js
```diff
@@ -0,0 +1,21 @@
...
+
+commands.mobileSetClipboard = async function (opts = {}) {
+  if (!util.hasValue(opts.content)) {
+    log.errorAndThrow(`The 'content' argument is mandatory`);
+  }
+  return await this.uiautomator2.jwproxy.command('/appium/clipboard/set', 'POST', opts);
+};
+
+commands.mobileGetClipboard = async function (opts = {}) {
+  return await this.uiautomator2.jwproxy.command('/appium/clipboard/get', 'POST', opts);
```
dpgraham: How the `/get` endpoint method uses `POST` and not `GET`?

jlipps: even so, could we include that as a url param like `/appium/clipboard/get/:contentType`?

mykola-mokhnach: I find POST more flexible %)

dpgraham: Can it not use query strings? (e.g.: `/appium/clipboard?contentType=<...>&...`

mykola-mokhnach:
I've decided to unify this with iOS and put the endpoint into base driver. Will update this PR as soon as base driver changes are available.

PR #154, opened 2018-03-28
allow isEmulator to be true when udid is set to an emulator

jlipps:
essentially copying what AndroidDriver does

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -203,7 +203,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
   }
 
   isEmulator () {
-    return !!this.opts.avd;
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -203,7 +203,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
   }
 
   isEmulator () {
-    return !!this.opts.avd;
+    return !!(this.opts.avd || /emulator/.test(this.opts.udid));
```
mykola-mokhnach: should the comparison be case-insensitive?

jlipps: i don't think so, that string is generated by google and is always lowercase AFAIK

jlipps:
published in 1.6.1

PR #156, opened 2018-03-31
Import logs broadcast via websocket feature

mykola-mokhnach:
This just imports what was implemented for android-driver in https://github.com/appium/appium-android-driver/pull/343


dpgraham:
Published to 1.8.0

PR #157, opened 2018-04-04
createSession should return session's capabilities

theoden9014:
`createSession` return the desired capabilities.
According to JSON Wire Protocol that `createSession` should return the actual session's capabilities.

https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#post-session

PR #160, opened 2018-04-11
Revert "createSession should return session's capabilities"

dpgraham:
Reverts appium/appium-uiautomator2-driver#157

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -102,14 +102,6 @@ commands.openNotifications = async function () {
   return await this.uiautomator2.jwproxy.command('/appium/device/open_notifications', 'POST', {});
 };
 
-commands.updateSettings = async function (settings) {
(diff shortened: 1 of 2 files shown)


imurchie:
I don't think this is necessary.

dpgraham:
This reverts it back to it's previous state? Are you saying reverting is unnecessary, or the original PR was unnecessary?

imurchie:
I think in this case the original commit was correct. There is more info in `this.caps` than `caps`. At least to my tests. But each driver has been implemented differently. I think we should go through the drivers and make sure that we are returning the complete info. 

As it stands base driver slaps a bunch of info into `this.caps` that is not in `caps`.  The sub classes then sometimes deal in `caps` and sometimes in `this.caps`.

imurchie:
`appium-xcuitest-driver`, `appium-uiautomator2-driver`, `appium-android-driver`, `appium-ios-driver` all seem ok in their current state, to my eye.

PR #161, opened 2018-04-13
Remove non-working endpoint call

mykola-mokhnach:
Looks like the wifi state changing helper inside uia2 server never worked properly, so we need to always use the one, which uses Appium Settings call with adb. See https://github.com/appium/ruby_lib_core/pull/72#issuecomment-380005798 for more details.


mykola-mokhnach:
@KazuCocoa FYI

PR #163, opened 2018-04-24
fix: Make tap by coordinates working

mykola-mokhnach:
It turns out that `parseTouch` helper inside android driver module always calculates absolute element coordinates instead of relative. Although these coordinates when being passed to `click` method of uia2 server are ignored and the tap action is simply translated to `element.click`. This change assumes that if coordinates are defined for the `tap` call then these coordinates are always absolute.


mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/10581

lib/commands/element.js
```diff
@@ -81,19 +82,25 @@ commands.getElementScreenshot = async function (element) {
...
 };
 
-commands.tap = async function (element, x = 0, y = 0, count = 1) {
+commands.tap = async function (elementId = null, x = null, y = null, count = 1) {
+  const params = {};
+  const areCoordinatesDefined = util.hasValue(x) && util.hasValue(y);
+  if (!util.hasValue(elementId) && !areCoordinatesDefined) {
+    throw new Error(`Either element to tap or both absolute coordinates should be defined`);
+  }
+  if (areCoordinatesDefined) {
```
vmaxim: Maybe it would be helpful to add log messages that ```elementId``` or ```x|y``` arguments will be ignored?

dpgraham:
This seems to have been forgotten about. @mykola-mokhnach is it still okay to merge this as-is?

mykola-mokhnach:
@dpgraham I believe https://github.com/appium/appium-android-driver/pull/355 should be merged and published first. As far as I remember @imurchie was concerned about the current implementation, but he unfortunately had no time to provide a better one.

PR #174, opened 2018-05-25
Force kill uiautomator process before server startup

mykola-mokhnach:
This might address https://github.com/appium/appium/issues/10749

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -286,7 +286,14 @@ class AndroidUiautomator2Driver extends BaseDriver {
     try {
       const pids = (await this.adb.getPIDsByName('uiautomator')).map((p) => `${p}`);
       if (!_.isEmpty(pids)) {
-        await this.adb.shell(['kill', '-9', ...pids]);
+        const isRoot = await this.adb.root();
+        try {
+          await this.adb.shell(['kill', '-9', ...pids]);
+        } finally {
+          if (isRoot) {
+            await this.adb.unroot();
+          }
+        }
       }
     } catch (err) {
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -283,10 +283,19 @@ class AndroidUiautomator2Driver extends BaseDriver {
       this.caps.appPackage = appInfo.appPackage;
     }
 
+    try {
+      const pids = (await this.adb.getPIDsByName('uiautomator')).map((p) => `${p}`);
+      if (!_.isEmpty(pids)) {
+        await this.adb.shell(['kill', '-9', ...pids]);
```
vmaxim: I thought device should be rooted to execute kill command. Am I wrong?

mykola-mokhnach: I asume it's only necessary if the process to which the signal is sent is running under `root` account, but uiautomator is usually running under `shell` account, so it shouldn't be the case.

vmaxim: Emu API 27
```
x:\>adb shell "ps | grep top" && adb shell kill -9 5884
shell         6528  5884    7832   3324 poll_schedule_timeout eb4fcaf0 S top
/system/bin/sh: kill: 5884: Permission denied
x:\>adb shell "ps | grep uiauto" && adb shell kill -9 7121
u10_a125      7121  1463 1037536  31372 ep_poll             0 S io.appium.uiautomator2.server
/system/bin/sh: kill: 7121: Operation not permitted
```

mykola-mokhnach:
Please do not merge this PR until it is verified on the client side

PR #186, opened 2018-06-26
Fix handling of autoWebview

imurchie:
We were having a default webview name of `WEBVIEW_0`, which never exists anymore. So use the same method as the `appium-android-driver`. This addresses https://github.com/appium/appium/issues/10784

Also fix the build, which has been failing on the adb port tests.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -328,9 +328,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
       const viewName = this.defaultWebviewName();
       const timeout = this.opts.autoWebviewTimeout || 2000;
       logger.info(`Setting auto webview to context '${viewName}' with timeout ${timeout}ms`);
-      await retryInterval(timeout / 500, 500, async () => {
-        await this.setContext(viewName);
-      });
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -324,8 +325,11 @@ class AndroidUiautomator2Driver extends BaseDriver {
     // if we want to immediately get into a webview, set our context
     // appropriately
     if (this.opts.autoWebview) {
-      await retryInterval(20, this.opts.autoWebviewTimeout || 2000, async () => {
-        await this.setContext(this.defaultWebviewName());
+      const viewName = this.defaultWebviewName();
+      const timeout = this.opts.autoWebviewTimeout || 2000;
+      logger.info(`Setting auto webview to context '${viewName}' with timeout ${timeout}ms`);
+      await retryInterval(timeout / 500, 500, async () => {
```
mykola-mokhnach: `await retryInterval(timeout / 500, 500, this.setContext(viewName))` ;)

imurchie:
Published in `appium-uiautomator2-driver@1.13.2`

PR #187, opened 2018-06-26
Stabilize the build

imurchie:
A few tweaks to make things run smoother.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -45,9 +45,9 @@ commands.getStrings = async function (language) {
   return this.apkStrings[language];
 };
 
-commands.getWindowSize = async function () {
+commands.getWindowSize = _.memoize(async function () {
   return await this.uiautomator2.jwproxy.command('/window/current/size', 'GET', {});
-};
+});
 
 // For W3C
 commands.getWindowRect = async function () {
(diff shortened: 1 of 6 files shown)


lib/commands/general.js
```diff
@@ -45,9 +45,9 @@ commands.getStrings = async function (language) {
   return this.apkStrings[language];
 };
 
-commands.getWindowSize = async function () {
+commands.getWindowSize = _.memoize(async function () {
```
mykola-mokhnach: @imurchie I assume this should be rather memoized per instance, but not per class

lib/commands/viewport.js
```diff
@@ -1,16 +1,17 @@
 import { imageUtil } from 'appium-support';
+import _ from 'lodash';
+
 
 let extensions = {}, commands = {};
 
-commands.getStatusBarHeight = async function () {
+commands.getStatusBarHeight = _.memoize(async function () {
```
mykola-mokhnach: same here

lib/commands/viewport.js
```diff
@@ -1,16 +1,17 @@
...
   const {statusBar} = await this.uiautomator2.jwproxy.command(`/appium/device/system_bars`, 'GET', {});
   return statusBar;
-};
+});
 
-// this function is memoized in the constructor
-commands.getDevicePixelRatio = async function () {
+commands.getDevicePixelRatio = _.memoize(async function () {
```
mykola-mokhnach: same here

PR #192, opened 2018-07-03
Add desired caps validation

mykola-mokhnach:
This is related to https://github.com/appium/appium-android-driver/pull/382

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -137,6 +137,13 @@ class AndroidUiautomator2Driver extends BaseDriver {
     }
   }
 
+  validateDesiredCaps (caps) {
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -137,6 +137,13 @@ class AndroidUiautomator2Driver extends BaseDriver {
     }
   }
 
+  validateDesiredCaps (caps) {
+    if (!super.validateDesiredCaps(caps)) {
```
jlipps: could be a bit simpler:

```js
return super.validateDesiredCaps(caps) && 
  androidHelpers.validateDesiredCaps(caps);
```

mykola-mokhnach: yep, changed

PR #193, opened 2018-07-05
support mjpeg stream screenshot via new mjpegScreenshotUrl cap

jlipps:
Implementation for the uiautomator2 driver of https://github.com/appium/appium-base-driver/pull/230, which this depends on.

### DIFF
--- a/lib/commands/screenshot.js
+++ b/lib/commands/screenshot.js
@@ -1,6 +1,16 @@
+import logger from '../logger';
+
 let commands = {};
 
 commands.getScreenshot = async function () {
+  if (this.mjpegStream) {
+    const data = await this.mjpegStream.lastChunkPNGBase64();
+    if (data) {
+      return data;
+    }
+    logger.warn("Tried to get screenshot from active MJPEG stream, but there " +
+                "was no data yet. Falling back to regular screenshot methods.");
+  }
   return await this.uiautomator2.jwproxy.command('/screenshot', 'GET');
 };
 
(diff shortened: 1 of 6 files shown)


jlipps:
Haven't gotten to tests for this yet; probably need to move the test mjpeg server to appium-test-support so we can use it in multiple places.

lib/commands/screenshot.js
```diff
@@ -1,6 +1,16 @@
+import logger from '../logger';
+
 let commands = {};
 
 commands.getScreenshot = async function () {
+  if (this.mjpegStream) {
+    const data = this.mjpegStream.lastChunkBase64;
+    if (data) {
+      return data;
```
mykola-mokhnach: I still don't think that it is a good idea to break the spec this way and send images in jpeg format instead of png

mykola-mokhnach: we could probably use jimp for this purpose

jlipps:
ok, PR updated to return PNG data and to include tests

jlipps:
package.json updated, let's see about CI

jlipps:
CI is not happy, however it's some weird error about node version and buffertools native module not compiling.

jlipps:
ah, @imurchie found the issue.

jlipps:
pushed @imurchie's android build fixes in this PR as well

jlipps:
ok, build passed.

jlipps:
published as `1.15.0`

PR #198, opened 2018-07-27
Fix the build

imurchie:
`appium-support` no longer automatically installs `mjpeg-consumer`, so install it in the Travis config.


jlipps:
will this need to be done other places as well? or did we actually use the functionality in this driver and that's why this problem cropped up?

imurchie:
I'm not sure where else this is actually used. But Greenkeeper will let us know, and I can fix.

PR #222, opened 2018-10-03
Fix strings extraction

mykola-mokhnach:
The original android-bootstrap component needs resource strings to be available as json on the destination device in order to "fix" element lookups. UIA2 server does not apply any fixes there, so there is no need to extract and upload strings json on session init. This change also allows to save a great amount of time. 

I'm going to delete the `/wd/hub/session/:sessionId/appium/app/strings` handler from UIA2 server, since it is a dummy implementation, which has absolutely no effect.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -35,7 +35,7 @@ commands.getStrings = async function (language) {
 
   // Clients require the resulting mapping to have both keys
   // and values of type string
-  const preprocessStringsMap = (mapping) => {
+  const preprocessStringsMap = function (mapping) {
     const result = {};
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -33,16 +33,50 @@ commands.getStrings = async function (language) {
     log.info(`No language specified, returning strings for: ${language}`);
   }
 
+  // Clients require the resulting mapping to have both keys
+  // and values of type string
+  const preprocessStringsMap = (mapping) => {
```
dpgraham: I prefer `function preprocessStringsMap () { ... }` so the name of the function shows up in the callstack if there's an error.

PR #242, opened 2018-11-06
add set hiddne api policy

KazuCocoa:
addressed in https://github.com/appium/appium-android-driver/pull/461#issuecomment-435773760

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -264,9 +264,16 @@ class AndroidUiautomator2Driver extends BaseDriver {
     // ADB instance
     this.adb = await androidHelpers.createADB(this.opts);
 
-    if (await this.adb.getApiLevel() < 21) {
+    const apiLevel = await this.adb.getApiLevel();
+
+    if (apiLevel < 21) {
       logger.errorAndThrow('UIAutomation2 is only supported since Android 5.0 (Lollipop). ' +
         'You could still use other supported backends in order to automate older Android versions.');
+    }
+
+    if (apiLevel >= 28) { // Android P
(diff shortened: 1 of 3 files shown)


lib/driver.js
```diff
@@ -264,11 +264,18 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
-    if (await this.adb.getApiLevel() < 21) {
+    const apiLevel = await this.adb.getApiLevel();
+
+    if (apiLevel < 21) {
       logger.errorAndThrow('UIAutomation2 is only supported since Android 5.0 (Lollipop). ' +
         'You could still use other supported backends in order to automate older Android versions.');
     }
 
+    if (apiLevel >= 28) { // Android P
+      logger.warn('Relaxing hidden api policy');
```
joshuapinter, line 275: Can we change this to `logger.info`? Two reasons: 

1. It matches the log level further down where you reset the API Policy to the default: 

    ```js
    if (await this.adb.getApiLevel() >= 28) { // Android P
      logger.info('Restoring hidden api policy to the device default configuration');
      await this.adb.setDefaultHiddenApiPolicy();
    }
    ```

2. It's not really a warning if it's always set and not something we can actually fix/remediate.

Plus, it shows up in our console logs, which are warning only and it's noisy: 

image

mykola-mokhnach: @joshuapinter Feel free to push a PR

joshuapinter: Excellent, that's what I wanted to hear.

source

joshuapinter: PR pushed at #396

PR #247, opened 2018-11-09
Another try at deflaking the keyboard tests

imurchie:
Still some flakiness with this, but it's much better on Travis.

### DIFF
--- a/test/functional/commands/keyboard/keyboard-e2e-specs.js
+++ b/test/functional/commands/keyboard/keyboard-e2e-specs.js
@@ -18,16 +18,29 @@ const PACKAGE = 'io.appium.android.apis';
 const TEXTFIELD_ACTIVITY = '.view.TextFields';
 const KEYEVENT_ACTIVITY = '.text.KeyEventText';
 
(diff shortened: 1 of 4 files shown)


test/functional/commands/keyboard/keyboard-e2e-specs.js
```diff
@@ -18,17 +18,30 @@ const PACKAGE = 'io.appium.android.apis';
...
   appActivity: TEXTFIELD_ACTIVITY
 });
 
-let defaultUnicodeCaps = Object.assign({}, defaultAsciiCaps, {
+const defaultUnicodeCaps = Object.assign({}, defaultAsciiCaps, {
   unicodeKeyboard: true,
   resetKeyboard: true
 });
 
+async function ensureUnlocked (driver) {
```
mykola-mokhnach: Is there a way to disable emulator lock while setting it on travis?

imurchie: Not that I have ever found, no.

PR #249, opened 2018-11-13
fix disabling animation

KazuCocoa:
As mentioned in https://github.com/appium/appium/pull/11640#issuecomment-438260477 , `--no-window-animation` doesn't disable animation duration correctly. It disables only window and transtion animation. Over Android8, it works correctly though.

Let me consider https://github.com/appium/appium-android-driver/pull/464 and https://github.com/appium/appium-espresso-driver as well a bit.


KazuCocoa:
Failing tests should be more stable after https://github.com/appium/appium-android-driver/pull/465
(disabling animations)

PR #262, opened 2018-12-01
Update server packages upgrade logic

mykola-mokhnach:
This change should improve driver startup performance and fix packages upgrade logic upon driver update


mykola-mokhnach:
Related to https://github.com/appium/appium/issues/11733

PR #265, opened 2018-12-11
Removed redundat UIAutomator processes killi during UIAutomator initialization

AnnaWyrwal:
Removed killing any existing UIAutomator2 processes running on device
during phase of UIAutomator2Server initialization, which is not required
at that point, due to possibility of re-installation apks. 

Processes are killed either way on the begging of session startup in function startSession in uiautomator2.js and I can't see the point in doing it twice.


imurchie:
Thanks for this improvement!

PR #280, opened 2019-01-11
Only try to cleanup uiautomator process if instrumentation crashes

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/11978

@mstrelex Would you mind to help with testing of this PR?


mstrelex:
Sure, ETA: Monday (don't have the device now).

mykola-mokhnach:
🍸

lib/driver.js
```diff
@@ -285,13 +285,13 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
 
-    // set up the modified UiAutomator2 server etc
-    await this.initUiAutomator2Server();
-
     // start an avd, set the language/locale, pick an emulator, etc...
     // TODO with multiple devices we'll need to parameterize this
     this.defaultIME = await helpers.initDevice(this.adb, this.opts);
 
+    // set up the modified UiAutomator2 server etc
+    await this.initUiAutomator2Server();
```
KazuCocoa: 👍

package.json
```diff
@@ -33,7 +33,6 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "adbkit": "^2.4.1",
```
KazuCocoa: 👍

mstrelex:
It works. thank you.

dpgraham:
@mykola-mokhnach Should I add this patch to 1.10?

mykola-mokhnach:
@dpgraham Yes, I would say this PR has a loot of goodness in it :)

lib/uiautomator2.js
```diff
@@ -99,116 +94,181 @@ class UiAutomator2Server {
...
+          } else if (!pmError) {
+            pmError = new Error('The instrumentation target is not listed by Package Manager');
+          }
         }
-        logger.debug(`Instrumentation '${INSTRUMENTATION_TARGET}' available`);
+        if (!isAmServiceAvailable) {
+          // https://travis-ci.org/appium/appium-uiautomator2-driver/jobs/478514097#L4806
+          amError = null;
+          try {
+            await this.adb.shell(['am']);
```
KazuCocoa: Is this command expected?

I failed test cases in ruby_lib_core because of this line.
https://gist.github.com/KazuCocoa/244e7cff066353be5dfc8e916d292c87

The result of `/Users/kazuaki/Library/Android/sdk/platform-tools/adb -P 5037 -s emulator-5554 shell am` which issued here was `255: command not found`. Then, the error caches and it becomes an error. We can see the list of command which is available `am` option.

Should we ignore the error only the result is `command not found`?

KazuCocoa: The result of command was:

```
$ /Users/kazuaki/Library/Android/sdk/platform-tools/adb -P 5037 -s emulator-5554 shell am
Activity manager (activity) commands:
  help
      Print this help text.
  start-activity [-D] [-N] [-W] [-P <FILE>] [--start-profiler <FILE>]
          [--sampling INTERVAL] [--streaming] [-R COUNT] [-S]
          [--track-allocation] [--user <USER_ID> | current] <INTENT>
      Start an Activity.  Options are:
      -D: enable debugging
      -N: enable native debugging
      -W: wait for launch to complete
      --start-profiler <FILE>: start profiler and send results to <FILE>
      --sampling INTERVAL: use sample profiling with INTERVAL microseconds
          between samples (use with --start-profiler)
      --streaming: stream the profiling output to the specified file
          (use with --start-profiler)
      -P <FILE>: like above, but profiling stops when app goes idle
      --attach-agent <agent>: attach the given agent before binding
      -R: repeat the activity launch <COUNT> times.  Prior to each repeat,
          the top activity will be finished.
      -S: force stop the target app before starting the activity
      --track-allocation: enable tracking of object allocations
      --user <USER_ID> | current: Specify which user to run as; if not
          specified then run as the current user.
      --stack <STACK_ID>: Specify into which stack should the activity be put.
...
```

The exit status was `255: command not found` in `$?` after the command

PR #285, opened 2019-01-30
Begin arranging for cron tests

imurchie:
Changes to the tests to accommodate using Sauce Labs
1. Stages instead of a build matrix
1. Pull driver destruction into a helper method

### DIFF
--- a/test/functional/helpers/session.js
+++ b/test/functional/helpers/session.js
@@ -2,10 +2,13 @@ import ADB from 'appium-adb';
 import { DEFAULT_HOST, DEFAULT_PORT } from '../../..';
 import logger from '../../../lib/logger';
 import wd from 'wd';
+import { retry } from 'asyncbox';
 import './mocha-scripts';
 
 
 const {SAUCE_RDC, SAUCE_EMUSIM, CLOUD, TRAVIS, CI} = process.env;
+
+const INIT_RETRIES = process.env.CI ? 2 : 1;
 
 function getPort () {
   if (SAUCE_EMUSIM || SAUCE_RDC) {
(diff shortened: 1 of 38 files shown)


test/functional/helpers/session.js
```diff
@@ -4,8 +4,30 @@ import logger from '../../../lib/logger';
 import wd from 'wd';
 
 
-async function initDriver (caps, adbPort) {
-  if (process.env.TRAVIS) {
+const {SAUCE_RDC, SAUCE_EMUSIM, CLOUD, TRAVIS, CI} = process.env;
+
+function getPort () {
+  if (SAUCE_EMUSIM || SAUCE_RDC) {
```
mykola-mokhnach: a ternary operator would also look good here

test/functional/helpers/session.js
```diff
@@ -35,4 +66,10 @@ async function initDriver (caps, adbPort) {
   return driver;
 }
 
-export { initDriver };
+async function deleteSession () {
+  try {
+    await driver.quit();
```
mykola-mokhnach: should we check if `driver` has value?

test/functional/helpers/session.js
```diff
@@ -14,13 +36,22 @@ async function initDriver (caps, adbPort) {
...
-  let driver = await wd.promiseChainRemote(DEFAULT_HOST, DEFAULT_PORT);
+  const host = getHost();
+  const port = getPort();
+  logger.debug(`Starting session on ${host}:${port}`);
+  driver = await wd.promiseChainRemote(host, port);
   await driver.init(caps);
 
   // In Travis, there is sometimes a popup
-  if (process.env.CI) {
+  if (CI && !CLOUD) {
```
mykola-mokhnach: is CI and TRAVIS the same thing? I don't really get the logic behind these conditions

test/functional/helpers/session.js
imurchie: Doesn't matter since we just ignore the error.

test/functional/helpers/session.js
imurchie: Travis sets a number of variables by default. https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

In this case, Travis emulators often get a popup that blocks automation. But when using Sauce, Travis is just running the test, not the emulator. So we need to check for it only on the condition where the emulator is running within Travis' infrastructure.

PR #296, opened 2019-02-21
fix no only test package case

KazuCocoa:
I faced the below case. Then, I must uninstall SERVER_PACKAGE_ID by manual to fix the issue. Without the uninstallation, never Appium becomes health.

- SERVER_TEST_PACKAGE_ID is not in a test device
- SERVER_PACKAGE_ID is in the test device
   - and SAME_VERSION_INSTALLED

Thus, I would like to enable `shouldInstallServerPackages` if the test device has no `test` package.
This case is not so common. So, I just enable it instead of separate the logic for SERVER_TEST_PACKAGE_ID and SERVER_PACKAGE_ID.

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -47,7 +47,7 @@ class UiAutomator2Server {
     let shouldInstallServerPackages = false;
     for (const {appId, appPath} of packagesInfo) {
       if (appId === SERVER_TEST_PACKAGE_ID) {
-        const isAppInstalled = false;// await this.adb.isAppInstalled(appId);
+        const isAppInstalled = await this.adb.isAppInstalled(appId);
 
         // There is no point in getting the state for test server,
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@imurchie Is it some problem on Travis that all e2e tests are failing?

imurchie:
Nope. This code is broken.

imurchie:
Seems like the server isn't being started, which is a code-loading problem. I have a build going to see if it fixes things: https://travis-ci.org/appium/appium-uiautomator2-driver/builds/496533907

KazuCocoa:
gotcha!

imurchie:
Seems it was the silent upgrade to mocha 6, since this package did not specify a mocha dependency. I've updated my PR and things seem to be working.

imurchie:
If you rebase with master things should work again.

PR #300, opened 2019-03-08
add timeoutcapname as uiautomator2ServerInstallTimeout to notify users about it

KazuCocoa:
depends on https://github.com/appium/appium-adb/pull/416


mykola-mokhnach:
I would also bump the minimum appium-adb component version in package.json

PR #317, opened 2019-04-12
Fix Travis build

imurchie:
Update Travis config, and add an alert id to the check for system alerts on session startup.


mykola-mokhnach:
@imurchie Do you ming applying similar changes to https://github.com/appium/appium-uiautomator2-server Travis setup?

imurchie:
Yes. That's next on my plate.

PR #325, opened 2019-06-26
Multi reporters

dpgraham:


umutuzgur:
@dpgraham As someone who doesn't know what multi reporters are, could you add a small description so I can understand better what this PR is doing?

dpgraham:
@umutuzgur yeah... mocha, the JS test runner, has different reporters you can use. The two we use are

`spec`: Spits out the test results to the command line and then at the end of the tests shows a summary of which tests failed (if any)

`junit-reporter`: Writes the test results out to an XML file that the CI can interpret. The CI presents the test results graphically and shows the history of test results

`mocha` doesn't allow multiple reporters to be run, by default. So the workaround is to use a reporter called `mocha-multi-reporter` which lets you run more than one reporter.

umutuzgur:
Got it. Thanks for explaining it

PR #337, opened 2019-09-09
feat: Make it possible to start UIA2 server without an app or a package id

mykola-mokhnach:


lib/driver.js
```diff
@@ -190,12 +190,14 @@ class AndroidUiautomator2Driver extends BaseDriver {
         // find and copy, or download and unzip an app url or path
         this.opts.app = await this.helpers.configureApp(this.opts.app, [APK_EXTENSION, APKS_EXTENSION]);
         await this.checkAppPresent();
-      } else if (this.appOnDevice) {
+      } else if (this.opts.appPackage) {
```
imurchie: Does this driver now take the `appPackage` from the `app` capability if it is there? Why not continue using `this.appOnDevice`?

mykola-mokhnach: `appOnDevice` was a blind copy paste from AndroidDriver, which is not even assigned anywhere in UIA2

mykola-mokhnach: > Does this driver now take the appPackage from the app capability if it is there?

It was always like that:

```
    // get appPackage et al from manifest if necessary
    const appInfo = await helpers.getLaunchInfo(this.adb, this.opts);
    // and get it onto our 'opts' object so we use it from now on
    Object.assign(this.opts, appInfo || {});
```

imurchie: Ah. Getters don't get assigned like functions do. Alas.

PR #342, opened 2019-10-22
fix: change errant uiautomation2 reference

imurchie:
Fixes https://github.com/appium/appium/issues/13471

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -272,7 +272,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
     const apiLevel = await this.adb.getApiLevel();
 
     if (apiLevel < 21) {
-      logger.errorAndThrow('UIAutomation2 is only supported since Android 5.0 (Lollipop). ' +
+      logger.errorAndThrow('UIAutomator2 is only supported since Android 5.0 (Lollipop). ' +
         'You could still use other supported backends in order to automate older Android versions.');
     }
 


mykola-mokhnach:
the actual automation backend in set before we know anything about the device, so we cannot automatically adjust the value depending on the target device if you are talking about that.

imurchie:
I don't understand. Your logs are for an attempt to run Appium using an automation technology that is not supported for the version of Android.

Setting the `automationName` capability to `'UiAutomator'` ought to work.

mindarchitect:
According to this
http://appium.io/docs/en/writing-running-appium/caps/index.html
There is no such automator as 'UiAutomator'.
And when trying to run with this automator it says
Original error: Could not find a driver for automationName 'UiAutomator' and platformName 'Android'. Please check your desired capabilities.
And in version 1.13 it was working with 'Appium' automator...
...
I have this overall feeling that latest versions are a bit bugy....

imurchie:
My apologies. The correct `automationName` is `'UiAutomator1'`. 

The default for Android was changed in Appium version [1.13.0](https://github.com/appium/appium/releases/tag/v1.13.0)

mindarchitect:
Now it runs. Thank you all for help. 
Update documentation please because this was exactly the case: I updated from 1.13 to 1.15 and tests stop to run because you just changed the automator name without any meaningful release notes...

Good luck!

imurchie:
I feel like it couldn't really get clearer in the release notes. https://github.com/appium/appium/releases/tag/v1.14.0

The documentation ought to be updated, I agree.

PR #346, opened 2019-11-05
feat: support new get events method by ensuring we don't try to proxy it

jlipps:


jlipps:
supports https://github.com/appium/appium-base-driver/pull/365

PR #348, opened 2019-11-16
feat: Get clipboard content via ADB if necessary on Android 10

mykola-mokhnach:
This is the part of https://github.com/appium/appium/issues/13466 implementation

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -11,6 +11,12 @@ commands.getPageSource = async function () {
   return await this.uiautomator2.jwproxy.command('/source', 'GET', {});
 };
 
+commands.getClipboard = async function () {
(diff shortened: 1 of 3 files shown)


lib/commands/general.js
```diff
@@ -11,6 +11,12 @@ commands.getPageSource = async function () {
   return await this.uiautomator2.jwproxy.command('/source', 'GET', {});
 };
 
+commands.getClipboard = async function () {
+  return (await this.adb.getApiLevel() < 29)
```
KazuCocoa: I noticed https://github.com/appium/appium-adb/pull/473/files#r347388767 was needed since the setting app already returned data as base64 encoded string.

After the change, this logic worked on android 10 👍

PR #365, opened 2020-02-21
feat: Add mobile command to type Unicode text

mykola-mokhnach:


mykola-mokhnach:
@jlipps The main difference is that sendKeys always overrides the existing field text. For example, if one needs to enter `st` into an input field containing `te` then it would be necessary to remember the current content, prefix the given content with it and only after that call `setText` (used in `sendKeys` call). This is not acceptable for cases where Android app under test reacts to each key press (we only expect to get notified with `st`, but it would be `test` instead).

There is also key actions W3C API, but it only supports ASCII chars and fails for Unicode (see the issue https://github.com/appium/appium/issues/13154).

So, the only way to achieve the desired input scenario would be the current endpoint.

mykola-mokhnach:
@KazuCocoa Yes, I did it deliberately, because this endpoint actually does not care about the actual focus, it just triggers the appropriate key codes and send them to the system thus emulating actual user interactions.

PR #369, opened 2020-03-05
fix: use allowOfflineDevices in tests

imurchie:
This should fix the build.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -27,10 +27,6 @@ const PORT_ALLOCATION_GUARD = new AsyncLock();
 // one of the ports above on the system to this port on the device.
 const DEVICE_PORT = 6790;
 
-// the initial access to ADB may fail because the device is not yet
-// in the correct state. a short wait clears this up
-const ADB_INITIAL_RETRIES = 5;
-
 // NO_PROXY contains the paths that we never want to proxy to UiAutomator2 server.
 // TODO:  Add the list of paths that we never want to proxy to UiAutomator2 server.
 // TODO: Need to segregate the paths better way using regular expressions wherever applicable.
@@ -299,10 +295,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
     // ADB instance
     this.adb = await androidHelpers.createADB(this.opts);
 
-    // the initial access to adb may fail for the device being offline
-    const apiLevel = await retryInterval(ADB_INITIAL_RETRIES, 100, async () => {
-      return await this.adb.getApiLevel();
-    });
+    const apiLevel = await this.adb.getApiLevel();
 
     if (apiLevel < 21) {
       logger.errorAndThrow('UIAutomator2 is only supported since Android 5.0 (Lollipop). ' +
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
@imurchie This PR would probably require the same Azure fix I did for https://github.com/appium/appium-espresso-driver/pull/550/commits/7c08dc7803101bf4c4b32d3cf3a671d71c93e599

lib/driver.js
```diff
@@ -287,15 +291,18 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
     this.opts.udid = udid;
     this.opts.emPort = emPort;
 
     // now that we know our java version and device info, we can create our
     // ADB instance
     this.adb = await androidHelpers.createADB(this.opts);
 
-    const apiLevel = await this.adb.getApiLevel();
+    // the initial access to adb may fail for the device being offline
+    const apiLevel = await retryInterval(ADB_INITIAL_RETRIES, 100, async () => {
```
mykola-mokhnach: should we only do the retries if the env.CI variable is set?

lib/driver.js
```diff
@@ -287,15 +291,18 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
+      return await this.adb.getApiLevel();
```
mykola-mokhnach: return is redundant

package.json
```diff
@@ -41,8 +41,8 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "appium-adb": "^7.20.0",
-    "appium-android-driver": "^4.25.3",
+    "appium-adb": "^7.22.0",
```
mykola-mokhnach: the recent version is 7.24.0

lib/driver.js
imurchie: I don't think we should be checking that inside actual code (as opposed to our tests).

This does not add any overhead, really, if it returns the first time.

lib/driver.js
imurchie: We want the API level to be returned from `retryInterval`, so we need to return it inside.

package.json
imurchie: Rebase error. I'll fix.

lib/driver.js
mykola-mokhnach: I mean `await retryInterval(ADB_INITIAL_RETRIES, 100, async () => await this.adb.getApiLevel())` would be enough

mykola-mokhnach: or even await retryInterval(ADB_INITIAL_RETRIES, 100, this.adb.getApiLevel.bind(this.adb)

lib/driver.js
mykola-mokhnach: this looks like a workaround to me. I would rather consider adding an option to `createADB` helper. This will address the issue for all android driver (since they all share the same helpers) and will make the stuff more flexible in general

test/functional/desired.js
```diff
@@ -15,10 +15,14 @@ const GENERIC_CAPS = {
   automationName: 'uiautomator2',
   adbExecTimeout: ADB_EXEC_TIMEOUT,
   allowOfflineDevices: true,
-  // TODO: newer build tools require Java9+
-  buildToolsVersion: '28.0.3',
 };
 
+if (process.env.CI && !process.env.TRAVIS) {
+  // TODO: newer build tools require Java9+
+  // which is not available on Azure
```
mykola-mokhnach: it is available, we just don't know how to switch :)

PR #378, opened 2020-03-29
feat: include sensorSet to mobile commands

vrunoa:
Running a few tests locally
Addressing https://github.com/appium/appium/issues/14116


KazuCocoa:
appium-uiautomator2-driver@1.48.0

PR #386, opened 2020-04-22
feat: Add settings related to UIAutomator2 mjpeg [screen] streaming feature + cleanup

jzaleski:
This change-set exposes the `mjpegServerPort` capability and starts / stops the ADB bind(s) / release(s) alongside the `{device,system}Port` bind / release.

This is related to PRs: https://github.com/appium/appium-uiautomator2-server/pull/340 and https://github.com/appium/appium/pull/14203 which were merged earlier this week.

Thank you for reviewing.

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -29,6 +29,9 @@ let uiautomatorCapConstraints = {
   systemPort: {
     isNumber: true
   },
+  mjpegServerPort: {
+    isNumber: true
+  },
   mjpegScreenshotUrl: {
     isString: true
   },
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -53,7 +53,10 @@ let uiautomatorCapConstraints = {
 };
 
 let desiredCapConstraints = {};
-Object.assign(desiredCapConstraints, uiautomatorCapConstraints,
-              commonCapConstraints);
+Object.assign(
+  desiredCapConstraints,
+  uiautomatorCapConstraints,
+  commonCapConstraints
+);
 
 export default desiredCapConstraints;
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -19,16 +19,13 @@ const helpers = Object.assign({}, uiautomator2Helpers, androidHelpers);
 // The range of ports we can use on the system for communicating to the
 // UiAutomator2 HTTP server on the device
 const DEVICE_PORT_RANGE = [8200, 8299];
-// The range of ports we can use on the system for communicating to the
-// UiAutomator2 MJPEG server on the device
-const MJPEG_SERVER_PORT_RANGE = [9200, 9299];
 
 // The guard is needed to avoid dynamic system port allocation conflicts for
 // parallel driver sessions
 const DEVICE_PORT_ALLOCATION_GUARD = new AsyncLock();
-// The guard is needed to avoid dynamic MJPEG server  port allocation conflicts
+// The guard is needed to avoid dynamic MJPEG server port allocation conflicts
 // for parallel driver sessions
-const MJPEG_SERVER_PORT_ALLOCATION_GUARD = new AsyncLock();
+const MJPEG_SERVER_DEVICE_PORT_ALLOCATION_GUARD = new AsyncLock();
 
 // This is the port that UiAutomator2 listens to on the device. We will forward
 // one of the ports above on the system to this port on the device.
@@ -36,7 +33,7 @@ const DEVICE_PORT = 6790;
 // This is the port that the UiAutomator2 MJPEG server listens to on the device.
 // We will forward one of the ports above on the system to this port on the
 // device.
-const MJPEG_SERVER_PORT = 7810;
+const MJPEG_SERVER_DEVICE_PORT = 7810;
 
 // NO_PROXY contains the paths that we never want to proxy to UiAutomator2 server.
 // TODO:  Add the list of paths that we never want to proxy to UiAutomator2 server.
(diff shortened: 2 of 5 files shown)


lib/driver.js
```diff
@@ -18,14 +18,25 @@ const helpers = Object.assign({}, uiautomator2Helpers, androidHelpers);
...
-// The guard is needed to avoid dynamic port allocation conflicts
+const DEVICE_PORT_RANGE = [8200, 8299];
+// The range of ports we can use on the system for communicating to the
+// UiAutomator2 MJPEG server on the device
+const MJPEG_SERVER_PORT_RANGE = [9200, 9299];
+
+// The guard is needed to avoid dynamic system port allocation conflicts for
+// parallel driver sessions
+const DEVICE_PORT_ALLOCATION_GUARD = new AsyncLock();
+// The guard is needed to avoid dynamic MJPEG server  port allocation conflicts
```
mykola-mokhnach, line 29: nit: duplicated space

lib/driver.js
```diff
@@ -281,10 +292,35 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
       async () => await this.adb.removePortForward(this.opts.systemPort));
   }
 
+  async allocateMjpegServerPort () {
+    await MJPEG_SERVER_PORT_ALLOCATION_GUARD.acquire(AndroidUiautomator2Driver.name, async () => {
+      try {
+        this.opts.mjpegServerPort = this.opts.mjpegServerPort || await findAPortNotInUse(
+          MJPEG_SERVER_PORT_RANGE[0], MJPEG_SERVER_PORT_RANGE[1]);
+      } catch (e) {
+        logger.errorAndThrow(
```
mykola-mokhnach, line 305: I'd rather print a warning und unset this.opts.mjpegServerPort capability to make it less intrusive

lib/driver.js
```diff
@@ -281,10 +292,35 @@ class AndroidUiautomator2Driver extends BaseDriver {
     if (!this.opts.systemPort) {
       return;
     }
-    await PORT_ALLOCATION_GUARD.acquire(AndroidUiautomator2Driver.name,
+    await DEVICE_PORT_ALLOCATION_GUARD.acquire(AndroidUiautomator2Driver.name,
       async () => await this.adb.removePortForward(this.opts.systemPort));
   }
 
+  async allocateMjpegServerPort () {
```
mykola-mokhnach, line 299: I don't see a reason to always automatically allocate this port if `this.opts.mjpegServerPort` is not provided explicitly. the automatically allocated port number number does not make much sense to an external client, since it is internal to the driver

lib/driver.js
jzaleski: Good catch. This will be resolved w/ the upcoming update(s).

lib/driver.js
jzaleski: This is resolved with the latest commit.

lib/driver.js
jzaleski: Makes sense. I have refactored how this bit works, please re-review w/ the latest commit.

jzaleski:
@mykola-mokhnach thanks for the suggestions. I refactored things a bit, please re-review.

lib/desired-caps.js
```diff
@@ -29,6 +29,21 @@ let uiautomatorCapConstraints = {
   systemPort: {
     isNumber: true
   },
+  mjpegServerPort: {
+    isNumber: true
+  },
+  mjpegServerFramerate: {
```
mykola-mokhnach, line 35: is is unnecessary to add these settings into capabilities. Appium supports initial settings setup by providing them in format of `settings[<setting_name>]` named capabilities.

lib/driver.js
```diff
@@ -281,10 +292,30 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
+    if (!this.opts.mjpegServerPort) {
+      return;
+    }
+    if (this.opts.mjpegServerPort < MJPEG_SERVER_PORT_RANGE[0] || this.opts.mjpegServerPort > MJPEG_SERVER_PORT_RANGE[1]) {
```
mykola-mokhnach, line 303: please remove this check. Let the user to decide the port number

jzaleski: OK.

jzaleski: This will be resolved in the upcoming changes.

lib/desired-caps.js
jzaleski: I was following the pattern for some of the other optional params listed in desired-caps.js -- good to know there is an alternative.

So I'd imagine I want to reference `this.settings.*` instead of `this.opts.*` down in the`initUiAutomator2Server` method?

mykola-mokhnach: actually, `this._settings`
See https://github.com/appium/appium/blob/6208275f493eb34a186bf9aae60befc8ae0dee09/lib/appium.js#L383
and 
https://github.com/appium/appium-base-driver/blob/ef9c11030e0563eec44daad587bd73c407d29122/lib/basedriver/device-settings.js#L75

for more details

jzaleski: It looks like `this.settings` is a `DeviceSettings` instance already in: lib/driver.js -- why don't I just toss the defaults in there?

mykola-mokhnach: yep, you're right. `this.settings` is correct. Sorry for the confusion

jzaleski: OK. Maybe I'm missing something w/ regards to the life-cycle here, so please help me understand 😸 

The intent was to set the initial configuration for the `initUiAutomator2Server` call, at that point there is no server to get settings from, right?

I was hoping to add the overrides into the _POST_ to: `/session` (e.g. session-creation) call if possible, doesn't that mean I need to toss them into `desiredCapabilities`?

mykola-mokhnach: I think it is fine to modify server settings after createSession call is completed. Users anyway are blocked while the endpoint is being called.

Just to summarize:
- We provide customized setting values in capabilities (as settings[..])
- We call createSession and start the server with default settings
- We set the desired settings right before the createSession call is completed
- We pass the control to the user

jzaleski: Thank you for the explanation. I had never used the `settings[<setting-name>]` concept before.. good to know that exists.

lib/driver.js
```diff
@@ -419,9 +447,8 @@ class AndroidUiautomator2Driver extends BaseDriver {
   }
 
   async initUiAutomator2Server () {
-    // now that we have package and activity, we can create an instance of
-    // uiautomator2 with the appropriate data
-    this.uiautomator2 = new UiAutomator2Server({
+    // broken out for readability
+    let uiautomator2Opts = {
```
mykola-mokhnach, line 451: const

lib/desired-caps.js
mykola-mokhnach: yep, I like to write code, but don't really like to keep the documentation up to date :(

jzaleski: This is all set.

lib/driver.js
jzaleski: This is done.

KazuCocoa:
appium-uiautomator2-driver@1.52.0

PR #387, opened 2020-04-24
chore: Use file-based locking for system port allocation

mykola-mokhnach:
AsyncLock only works for sessions executed in scope of the same Appium process, but is useless if sessions are started on behalf of parallel processes. file-based lock resolves that problem, since it can also be shared between processes


jlipps:
Nice, glad to see the file locking utility having more ... utility!

PR #389, opened 2020-04-27
feat: Add an option to control intent wait for a started URI

shibupanda:
Added condition to handle waitForLaunch capability to call concerned startUrl function.
Which will fix below issue.
https://github.com/appium/appium/issues/14180
It's depends on below PR
https://github.com/appium/appium-adb/pull/520

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -232,17 +232,17 @@ commands.mobileViewportScreenshot = async function () {
 };
 
 commands.setUrl = async function (url) {
-  if (!this.opts.waitForLaunch) {
-    await this.adb.startUri(url, this.opts.appPackage, this.opts.waitForLaunch);
-  } else {
-    await this.adb.startUri(url, this.opts.appPackage);
-  }
+  await this.adb.startUri(url, this.opts.appPackage);
 };
 
 commands.mobileDeepLink = async function (opts = {}) {
-  const {url, package: pkg} = opts;
-  return await this.adb.startUri(url, pkg);
+  const {
+    url,
+    package: pkg
+  } = opts;
+  return await this.adb.startUri(url, pkg, opts);
 };
+
 
 commands.openNotifications = async function () {
   return await this.uiautomator2.jwproxy.command('/appium/device/open_notifications', 'POST', {});
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -232,7 +232,11 @@ commands.mobileViewportScreenshot = async function () {
 };
 
 commands.setUrl = async function (url) {
-  await this.adb.startUri(url, this.opts.appPackage);
+  if (!this.opts.waitForLaunch) {
+    await this.adb.startUri(url, this.opts.appPackage, this.opts.waitForLaunch);
+  } else {
+    await this.adb.startUri(url, this.opts.appPackage);
+  }
 };
 
 commands.mobileDeepLink = async function (opts = {}) {
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -238,11 +238,11 @@ commands.setUrl = async function (url) {
 commands.mobileDeepLink = async function (opts = {}) {
   const {
     url,
-    package: pkg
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -236,10 +236,14 @@ commands.setUrl = async function (url) {
 };
 
 commands.mobileDeepLink = async function (opts = {}) {
-  const {url, package: pkg} = opts;
-  return await this.adb.startUri(url, pkg);
+  const {
+    url,
+    package: pkg
```
mykola-mokhnach, line 241: Please destruct waitForAppLaunch  explicitly from opts and add a doctring to this method

lib/commands/general.js
```diff
@@ -236,10 +236,14 @@ commands.setUrl = async function (url) {
...
+  } = opts;
+  return await this.adb.startUri(url, pkg, opts);
 };
 
+
```
mykola-mokhnach, line 246: no need for an empty line

package.json
```diff
@@ -41,7 +41,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "appium-adb": "^7.29.0",
+    "appium-adb": "^7.30.0",
```
mykola-mokhnach, line 44: 👍

lib/commands/general.js
shibupanda: Is it something like below?
commands.mobileDeepLink = async function (opts = {}) {
  const {
    url,
    package: pkg,
    waitForLaunch: wait,
  } = opts;
  return await this.adb.startUri(url, pkg, wait);
};

mykola-mokhnach: something like

```
const {
  url,
  package,
  waitForLaunch,
} = opts;
return await this.adb.startUri(url, package, {waitForLaunch});
```

shibupanda: Yes this also worked for me. Let me do it.
And docstring was not there you want me to add that too?

mykola-mokhnach: yes, I know. But we want to make it better, don't we?

shibupanda: Very true @mykola-mokhnach.

mykola-mokhnach:
@shibupanda Do you have time to add the docstring for the `mobileDeepLink` method?

shibupanda:
@mykola-mokhnach Sorry for the delay I stuck some where so will add it in this weekend or I can create a different pull request for it.

mykola-mokhnach:
np, lets do it on the weekend

lib/commands/general.js
```diff
@@ -234,7 +234,18 @@ commands.mobileViewportScreenshot = async function () {
 commands.setUrl = async function (url) {
   await this.adb.startUri(url, this.opts.appPackage);
 };
+/**
+ * @typedef {object} DeepLinkOpts
+ * @property {?string} url - The name of URL to start.
```
mykola-mokhnach, line 239: the type would be `{!string}` since this property is required

lib/commands/general.js
```diff
@@ -234,7 +234,18 @@ commands.mobileViewportScreenshot = async function () {
...
+ * @property {?string} package - The name of the package to start the URI with.
```
mykola-mokhnach, line 240: same here

lib/commands/general.js
shibupanda: Oh! Yes.

shibupanda:
This needs to be update [here](https://github.com/appium/appium/blob/17004739b90c18caac5b89c1afbe691a3f8a766b/commands-yml/commands/mobile-command.yml#L90) I guess.

shibupanda:
> This needs to be update [here](https://github.com/appium/appium/blob/17004739b90c18caac5b89c1afbe691a3f8a766b/commands-yml/commands/mobile-command.yml#L90) I guess.

Shall I proposed for the changes?

mykola-mokhnach:
simply create a PR

PR #396, opened 2020-05-26
fix: Change API Policy log level from warn to info

joshuapinter:
Two reasons: 

1. It matches the log level further down where you reset the API Policy to the default: 

    ```js
    if (await this.adb.getApiLevel() >= 28) { // Android P
      logger.info('Restoring hidden api policy to the device default configuration');
      await this.adb.setDefaultHiddenApiPolicy();
    }
    ```

2. It's not really a warning if it's always set and not something we can actually fix/remediate.

Plus, it shows up in our console logs, which are warning only and it's noisy: 

image

See https://github.com/appium/appium-uiautomator2-driver/pull/242#discussion_r430560813 for original discussion.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -321,7 +321,7 @@ class AndroidUiautomator2Driver extends BaseDriver {
     }
 
     if (apiLevel >= 28) { // Android P
-      logger.warn('Relaxing hidden api policy');
+      logger.info('Relaxing hidden api policy');
       await this.adb.setHiddenApiPolicy('1', !!this.opts.ignoreHiddenApiPolicyError);
     }
 


KazuCocoa:
`appium-uiautomator2-driver@1.54.2` has this, thanks

joshuapinter:
Beauty, thanks!

joshuapinter:
Quick question: I've pulled version 1.15.2 and verified that our appium `--log-level` is set to `warn:debug` (`warn` for console` and `debug` for log file) but I'm still seeing this message in the console log: 

```
> cntral_mobile@ appium /Users/me/Development/cntral_mobile
> appium --log __tests__/appium/logs/appium.log --log-level warn:debug

2020-05-27 12:53:37:984 - [UiAutomator2] Relaxing hidden api policy
```

Any ideas?

mykola-mokhnach:
>  I've pulled version 1.15.2 

Did you only pull that version or appium@beta?

joshuapinter:
Sorry, I meant 1.54.2. 

And I only pulled that version of `appium-uiautomator2-driver`: 

```json
"appium-uiautomator2-driver": "1.54.2",
```

I didn't tough the `appium` package. Does the `appium` package use its "own" version of `appium-uiautomator2-driver`?

mykola-mokhnach:
yes, there are cross-dependencies and it is usually necessary to upgrade the whole package instead of single module reinstall

PR #404, opened 2020-06-14
feat: Expose ViewPortRect mobile command.

itaibh:
In order to be able to locate native elements correctly, one must be able to tell the viewport rectangle, in addition to getting the viewport screenshot. That's because native elements return with locations relative to the device, not the viewport.
So now, once we have a native element's coordinates, we can subtract the viewport's origin coordinate and the result will be the native element's coordinates relative to the viewport.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -110,6 +110,7 @@ extensions.executeMobile = async function (mobileCommand, opts = {}) {
     scrollBackTo: 'mobileScrollBackTo',
     scroll: 'mobileScroll',
     viewportScreenshot: 'mobileViewportScreenshot',
+    viewportRect: 'mobileViewPortRect',
 
     deepLink: 'mobileDeepLink',
 
@@ -232,6 +233,10 @@ commands.mobileScroll = async function (opts = {}) {
 
 commands.mobileViewportScreenshot = async function () {
   return await this.getViewportScreenshot();
+};
+
+commands.mobileViewPortRect = async function () {
+  return await this.getViewPortRect();
 };
 
 commands.setUrl = async function (url) {
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -235,6 +235,18 @@ commands.mobileViewportScreenshot = async function () {
   return await this.getViewportScreenshot();
 };
 
+/**
+ * @typedef {object} Rectangle
+ * @property {number} left - The left coordinate of the Rectangle.
+ * @property {number} top - The top coordinate of the Rectangle.
+ * @property {number} width - The width of Rectangle.
+ * @property {number} height - The height of Rectangle.
+ */
+
+/**
+ * Returns the viewport coordinates.
+ * @returns {Rectangle} The viewport coordinates.
+ */
 commands.mobileViewPortRect = async function mobileViewPortRect () {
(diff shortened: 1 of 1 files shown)


lib/commands/general.js
```diff
@@ -234,6 +235,10 @@ commands.mobileViewportScreenshot = async function () {
   return await this.getViewportScreenshot();
 };
 
+commands.mobileViewPortRect = async function () {
```
mykola-mokhnach, line 238: Please add a docstring on the method about what it is supposed to return

mykola-mokhnach, line 238: Please add a name to the function - this way it's easier to debug such calls

itaibh: I copied `mobileViewportScreenshot` implementation in this file to `mobileViewPortRect`, which isn't documented with the return value, and I didn't see any other method documented with return value. The actual `commands.getViewPortRect` method was already implemented in the `viewport.js` file. So I don't understand what should I do.

itaibh: What do you mean? How do I "add a name to the function"? It's name is `mobileViewPortRect`, isn't it?

mykola-mokhnach:
```
async function () { -> async function mobileViewPortRect () { 
```

mykola-mokhnach: Here is an example: https://github.com/appium/appium-uiautomator2-driver/pull/404/files#diff-046460f6a24bb8c575cc6d5f55189736R217

itaibh: You mean like in the last method in that file? `mobileType`? This is the only method implemented that way...

itaibh: You mean like in the last method in that file? `mobileType`? This is the only method implemented that way...

KazuCocoa:
It would be great to add the command in https://github.com/appium/appium/blob/master/commands-yml/commands/mobile-command.yml later

itaibh:
> It would be great to add the command in https://github.com/appium/appium/blob/master/commands-yml/commands/mobile-command.yml later

@KazuCocoa done in https://github.com/appium/appium/pull/14450

lib/commands/general.js
mykola-mokhnach: yes, similar to that one

itaibh: Done in a recent commit https://github.com/appium/appium-uiautomator2-driver/pull/404/commits/83dfefed16382e993d31b26f4efd6ccf1eecee42 from yesterday.

itaibh:
I believe I followed and implemented all the requested changes. Did I miss anything? Any other changes you want me to implement? Or can this PR be merged and closed?

mykola-mokhnach:
https://github.com/appium/appium/pull/14450/files#diff-29516863f13863f4f028f30051fe44a2R96 Does not actually describe which properties to expect from the returned object and what it is. That is why I've requested to add a docstring. For a client it is important to know that the returned data structure is a map with ... properties.

lib/commands/general.js
itaibh: I hope commit https://github.com/appium/appium-uiautomator2-driver/pull/404/commits/63d0703948b9bf5913f0c354db50018cb84103b5 was done correctly.

PR #410, opened 2020-09-04
feat: native CSS selector for Android UiAutomator

dpgraham:
The CSS selector is going unused in all Appium Drivers 😢 . Why not put it to use?

This PR adds support for a native CSS selector strategy. Instead of using CSS to select HTML elements, it parses CSS selectors and translate the selector to a native `-android uiautomator` selector.

## Example

`android.widget.TextView[description="Some description"]` 

translates to 

`new UiSelector().className("android.widget.TextView").description("Some description")` 

ie:

`elementByCss('android.widget.TextView[description="Some description"]')`

does the same thing as

`elementBy('-android uiautomator', 'new UiSelector().className("android.widget.TextView").description("Some description")')`

_many more examples in **css-converter-specs.js** and **by-css-e2e-specs.js**_

## Versatile
This selector can be used to replace every other selector:

* find by id: 
`elementsById("someResourceID")` --> 
`elementsByCss("#someResourceID")`

* find by class name: 
`elementsByClassName("android.widget.TextView")` -->
`elementsByCss("android.widget.TextView")`

* find by accessibility id: 
`elementsByAccessibilityId("Some Content Description")` -->
`elementsByCss('*[description="Some Content Description"]')`

* find by xpath: 
`elementsByXpath("//android.widget.TextView[@description='Accessibility']")` --> `elementsByCss("android.widget.TextView[description='Accessibility']")`

## Performance
Unlike Xpath, this selector doesn't require a document to be generated that the selector is applied to. It just parses the CSS selector and translates it to an analagous UiAutomator selector.

## How is it done?
* css is parsed by css-converter.js which uses the https://www.npmjs.com/package/css-selector-parser library
* the CSS object is parsed and converted to a UiSelector expression that is passed to UiAutomator2 Server as `-android uiautomator` selector

## How is it different from CSS selectors being applied to HTML?
* Throws exceptions if unsupported attributes are given. This is so that it's more helpful for users so they can see when they're using a bad attribute
* The descendant operators (`>`, `~`, ...) are unsupported.
* Class names:
  * For HTML CSS selectors, a selector like `p.classOne.classTwo` means that it's matching an HTML `<p>` element that has `classOne` and `classTwo` as class attributes (order doesn't matter). 
  * For this UiAutomator2 CSS selector, a selector like `android.widget.TextView` means that it's matching a widget with the fully qualified Java class name `android.widget.TextView`

## Next Step
Support this in iOS with class chain and in Espresso with ViewMatchers and DataMatchers


lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  log.errorAndThrow(`Could not parse '${css.name}=${css.value}'. '${css.name}' must be true, false or empty`);
+}
+
+/**
+ * Get the canonical form of a CSS attribute name
+ *
+ * Converts to lowercase and if an attribute name is an alias for something else, return
+ * what it is an alias for
+ *
+ * @param {*} css CSS object
```
mykola-mokhnach, line 69: Please add return description

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  ['description', [
+    'content-description', 'content-desc',
+    'desc', 'accessibility-id',
+  ]],
+  ['index', ['nth-child']],
+];
+
+/**
+ * Convert hyphen separated word to snake case
+ * @param {String} str
```
mykola-mokhnach, line 42: Please add return description. Same below

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+function toSnakeCase (str) {
+  const tokens = str.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1).toLowerCase());
+  let out = tokens.join('');
+  return out.charAt(0).toLowerCase() + out.slice(1);
+}
+
+/**
+ * Get the boolean from a CSS object. If empty, return true. If not true/false/empty, throw exception
+ *
+ * @param {*} css CSS object
```
mykola-mokhnach, line 53: Please add typedef description of the CSS object

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  ]],
+  ['index', ['nth-child']],
+];
+
+/**
+ * Convert hyphen separated word to snake case
+ * @param {String} str
+ */
+function toSnakeCase (str) {
+  const tokens = str.split('-').map((str) => str.charAt(0).toUpperCase() + str.slice(1).toLowerCase());
```
mykola-mokhnach, line 45: this will throw an exception if str.length is 0

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  return out.charAt(0).toLowerCase() + out.slice(1);
+}
+
+/**
+ * Get the boolean from a CSS object. If empty, return true. If not true/false/empty, throw exception
+ *
+ * @param {*} css CSS object
+ */
+function assertGetBool (css) {
+  const val = css.value?.toLowerCase() || 'true';
```
mykola-mokhnach, line 56: why `|| 'true'` ?

dpgraham: @mykola-mokhnach good question...

Answer with an example:
* In css `inputs[checked]` implies that you're matching inputs that are checked (ie: where `checked=true`. So if the value is omited, it's implied that `checked=true`

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+
+const CssConverter = {};
+
+const parser = new CssSelectorParser();
+parser.registerSelectorPseudos('has');
+parser.registerNestingOperators('>', '+', '~');
+parser.registerAttrEqualityMods('^', '$', '*', '~');
+parser.enableSubstitutes();
+
+const booleanAttrs = [
```
mykola-mokhnach, line 12: constant names are usually in uppercase

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+function getWordMatcherRegex (word) {
+  return `\\b(\\w*${escapeRegexLiterals(word)}\\w*)\\b`;
+}
+
+/**
+ * Add escapes to regex literals
+ *
+ * @param {*} str
+ */
+function escapeRegexLiterals (str) {
```
mykola-mokhnach, line 103: lodash already has a helper to escape regex

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+ * @param {string} word
+ */
+function getWordMatcherRegex (word) {
+  return `\\b(\\w*${escapeRegexLiterals(word)}\\w*)\\b`;
+}
+
+/**
+ * Add escapes to regex literals
+ *
+ * @param {*} str
```
mykola-mokhnach, line 101: `{string}`

lib/css-converter.js
dpgraham: * `inputs[checked=true]` -> `checked(true)`
* `inputs[checked=false]` -> `checked(false)`
* `inputs[checked]` -> `checked(true)`

lib/css-converter.js
mykola-mokhnach: same for others

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * Convert a CSS attribute into a UiSelector method call
+ * @param {*} cssAttr CSS attribute object
+ */
+function parseAttr (cssAttr) {
+  if (cssAttr.valueType && cssAttr.valueType !== 'string') {
+    log.errorAndThrow(`Could not parse '${cssAttr.name}=${cssAttr.value}'.` +
+      `Unsupported attribute type '${cssAttr.valueType}'`);
```
mykola-mokhnach, line 127: We also could enumerate which types we do support

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  if (strAttrs.includes(attrName)) {
+    let value = cssAttr.value || '';
+    if (attrName === 'resource-id') {
+      value = prependAndroidId(value);
+    }
+    if (value === '') {
+      return `.${methodName}Matches("")`;
+    } else {
```
mykola-mokhnach, line 142: else is redundant after return

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  }
+  const attrName = assertGetAttrName(cssAttr);
+  const methodName = toSnakeCase(attrName);
+  if (booleanAttrs.includes(attrName)) {
+    return `.${methodName}(${assertGetBool(cssAttr)})`;
+  }
+
+  if (strAttrs.includes(attrName)) {
+    let value = cssAttr.value || '';
+    if (attrName === 'resource-id') {
```
mykola-mokhnach, line 137: 'resource-id' could be moved to a constant

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+          if (['description', 'text'].includes(attrName)) {
+            return `.${methodName}StartsWith("${value}")`;
+          }
+          return `.${methodName}Matches("^${escapeRegexLiterals(value)}")`;
+        case '$=':
+          return `.${methodName}Matches("${escapeRegexLiterals(value)}$")`;
+        case '~=':
+          return `.${methodName}Matches("${getWordMatcherRegex(value)}")`;
+        default:
+          log.errorAndThrow(`Unsupported CSS attribute operator '${cssAttr.operator}'`);
```
mykola-mokhnach, line 161: Please enumerate which attributes are supported

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  }
+}
+
+/**
+ * Convert a CSS pseudo class to a UiSelector
+ * @param {*} cssPseudo CSS Pseudo class
+ */
+function parsePseudo (cssPseudo) {
+  if (cssPseudo.valueType && cssPseudo.valueType !== 'string') {
+    log.errorAndThrow(`Could not parse '${cssPseudo.name}=${cssPseudo.value}'. ` +
```
mykola-mokhnach, line 173: We also could enumerate which types we do support

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  if (numericAttrs.includes(pseudoName)) {
+    return `.${pseudoName}(${cssPseudo.value})`;
+  }
+}
+
+/**
+ * Convert a CSS rule to a UiSelector
+ * @param {*} cssRule CSS rule definition
+ */
+function parseCssRule (cssRule) {
```
mykola-mokhnach, line 192: I think it makes sense to have a class for such transformation rather than a set of functions.

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+  let uiAutomatorSelector = 'new UiSelector()';
```
mykola-mokhnach, line 193: Did you think about having `xpath` instead of UiSelector? With xpath we could also easily migrate this functionality into xcuitest or Espresso. Although, UiSelector is only specific to UIAutomator

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+    case 'rule':
+      return parseCssRule(css);
+    case 'ruleSet':
+      return parseCssObject(css.rule);
+    case 'selectors':
+      return css.selectors.map((selector) => parseCssObject(selector)).join('; ');
+
+    default:
+      // This is never reachable, but if it ever is do this.
+      throw new Error(`UiAutomator does not support '${css.type}' css`);
```
mykola-mokhnach, line 237: Please add which rules are supported

lib/css-converter.js
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Convert a CSS selector to a UiAutomator2 selector
+ * @param {string} cssSelector CSS Selector
+ */
+CssConverter.toUiAutomatorSelector = function toUiAutomatorSelector (cssSelector) {
+  let cssObj;
+  try {
+    cssObj = parser.parse(cssSelector);
+  } catch (e) {
+    log.errorAndThrow(`Could not parse CSS. Reason: '${e}'`);
```
mykola-mokhnach, line 250: I would rather add the actual selector string into the error message

test/functional/commands/find/by-css-e2e-specs.js
```diff
@@ -0,0 +1,78 @@
+import chai from 'chai';
```
mykola-mokhnach, line 1: I like the tests. It would also make sense to add a documentation to this feature

lib/css-converter.js
mykola-mokhnach: Please add the explanation above as a comment

test/functional/commands/find/by-css-e2e-specs.js
dpgraham: @mykola-mokhnach definitely. I'll add something to the appium docs and then probably an AppiumPro article.

lib/css-converter.js
dpgraham: @mykola-mokhnach I chose `UiSelector` because I want this to be a selector that gets native performance. I want to apply this same code to iOS and have it translate to `class chain` and to Espresso and have it translate to the ViewMatcher and DataMatcher selectors.

lib/css-converter.js
dpgraham: @mykola-mokhnach I prefer the procedural approach myself, but that's a much larger debate.

lib/css-converter.js
dpgraham: Found this: https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript ... copied and pasted. I couldn't find lodash version of this.

dpgraham:
Suggestions implemented.

test/functional/commands/find/by-css-e2e-specs.js
dpgraham: I'm going to do the iOS and Android-Espresso versions of this next. The Espresso version is going to be awesome... a hybrid of datamatchers and viewmatchers.

lib/css-converter.js
mykola-mokhnach: https://lodash.com/docs/4.17.15#escapeRegExp

lib/css-converter.js
mykola-mokhnach: having a class could help to avoid passing the css object (and possibly many other shared args) to all the functions

mykola-mokhnach: also, this could be refactored to have a main class that performs css expression validation and tokenization and then passing a tranformer class/function to it. This class/function could be responsible for the expression tranformation itself to. let say UiSelector or class chain expression. This way we could extract the common part of the code and possibly reuse it by moving to appium-support module. So, only the driver-specific transformation method would stay in the driver source.

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+  'description', 'resource-id', 'text', 'class-name', 'package-name'
+];
+
+const ALL_ATTRS = [
+  ...BOOLEAN_ATTRS,
+  ...NUMERIC_ATTRS,
+  ...STR_ATTRS,
+];
+
+const attributeAliases = [
```
mykola-mokhnach, line 31: ATTRIBUTE_ALIASES

mykola-mokhnach: this could be a map

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+const RESOURCE_ID = 'resource-id';
+
+/**
+ * Convert hyphen separated word to snake case
+ *
+ * @param {string} str
+ * @returns {string} The hyphen separated word translated to snake case
+ */
+function toSnakeCase (str) {
+  const tokens = str.split('-').map((str) => str.length > 0 ? str.charAt(0).toUpperCase() + str.slice(1).toLowerCase() : '');
```
mykola-mokhnach, line 49: how about
```
if (!str) {
  return str
}
```

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+  log.errorAndThrow(`Could not parse '${css.name}=${css.value}'. '${css.name}' must be true, false or empty`);
+}
+
+/**
+ * Get the canonical form of a CSS attribute name
+ *
+ * Converts to lowercase and if an attribute name is an alias for something else, return
+ * what it is an alias for
+ *
+ * @param {Object} css CSS object
```
mykola-mokhnach, line 74: I would rather prefer more specific definition of the CSS object with properties. Something like https://github.com/appium/appium-uiautomator2-driver/blob/f304d2c166c94bb721a884d2c96b5fda80d98a0a/lib/commands/alert.js#L8

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+function prependAndroidId (str) {
+  if (!str.startsWith('android:id/')) {
+    return `android:id/${str}`;
+  }
+  return str;
+}
+
+/**
+ * @typedef {Object} CssAttr
+ * @param {?string} valueType Type of attribute (must be string or empty)
```
mykola-mokhnach, line 132: property

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+  }
+  if (cssRule.rule) {
+    uiAutomatorSelector += `.childSelector(${parseCssRule(cssRule.rule)})`;
+  }
+  return uiAutomatorSelector;
+}
+
+/**
+ * @typedef {Object} CssObject
+ * @param {Object} type Type of CSS object. 'rule', 'ruleset' or 'selectors'
```
mykola-mokhnach, line 271: string?

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+const ATTRIBUTE_ALIASES = [
+  ['resource-id', ['id']],
+  ['description', [
+    'content-description', 'content-desc',
+    'desc', 'accessibility-id',
+  ]],
+  ['index', ['nth-child']],
+];
+
+const RESOURCE_ID = 'resource-id';
```
mykola-mokhnach, line 40: this constant could be reused in other constants above

lib/css-converter.js
```diff
@@ -0,0 +1,311 @@
...
+function assertGetAttrName (css) {
+  const attrName = css.name.toLowerCase();
+
+  // Check if it's supported and if it is, return it
+  if (ALL_ATTRS.includes(attrName)) {
+    return attrName.toLowerCase();
+  }
+
+  // If attrName is an alias for something else, return that
+  for (let [officialAttr, aliasAttrs] of ATTRIBUTE_ALIASES) {
```
mykola-mokhnach, line 86: `const [officialAttr, aliasAttrs]`

lib/css-converter.js
dpgraham: I chose this approach because of my personal preference ... I find the testing simpler and I personally lean towards avoiding the cognitive overhead of thinking about "state". That is it to say, when I'm writing and troubleshooting functions, I prefer the functions to be "pure" so that I only have to think about the inputs (the args) and the output and that I don't have to think about the state of the object under test.

I still believe there are many cases where using classes/objects is the right approach. I won't be advocating that we port Appium to Haskell any time soon. I like that Javascript affords the choice between functional programming and object-oriented :)

lib/css-converter.js
dpgraham: I considered doing it that way, but then I would have had to do `_.toPairs(...)`, which would have translated it to this format. Defining it as a two-column matrix cuts out the middle-man.

lib/css-converter.js
mykola-mokhnach: It is fine to use Haskell-based approach. For me it's more important to have it properly designed for architectural perspective. E.g. that we could easily separate different entities (tokenization, validation and transformation) and then implement the feature in different drivers without copying the same parts of code here and there

dpgraham: For this feature, my plan is to not share any code between the drivers and just copy-paste + refactor.

Normally this is terrible coding practice, but I'm making an exception because this is based on the CSS selector standard, which isn't going to be introducing breaking changes any time soon and it's very unlikely that we'll ever need to make alterations to the code that would have to be ported to all drivers. I would rather each driver do it's own implementation.

test/unit/css-converter-specs.js
```diff
@@ -0,0 +1,67 @@
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import CssConverter from '../../lib/css-converter';
+
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('css-converter.js', function () {
+  describe('simple cases', function () {
+    const simpleCases = [
```
KazuCocoa, line 10: awesome

lib/css-converter.js
```diff
@@ -0,0 +1,307 @@
...
+ * Convert hyphen separated word to snake case
+ *
+ * @param {string} str
+ * @returns {string} The hyphen separated word translated to snake case
+ */
+function toSnakeCase (str) {
+  if (!str) {
+    return '';
+  }
+  const tokens = str.split('-').map((str) => str.length > 0 ? str.charAt(0).toUpperCase() + str.slice(1).toLowerCase() : '');
```
mykola-mokhnach, line 53: str.length check is now not needed here

lib/css-converter.js
```diff
@@ -0,0 +1,307 @@
...
+  let out = tokens.join('');
```
mykola-mokhnach, line 54: const

lib/css-converter.js
```diff
@@ -0,0 +1,307 @@
...
+    return attrName.toLowerCase();
+  }
+
+  // If attrName is an alias for something else, return that
+  for (const [officialAttr, aliasAttrs] of ATTRIBUTE_ALIASES) {
+    if (aliasAttrs.includes(attrName)) {
+      return officialAttr;
+    }
+  }
+  log.errorAndThrow(`'${attrName}' is not a valid attribute. ` +
```
mykola-mokhnach, line 101: I would rather throw a more specific Locator syntax error everywhere: https://github.com/appium/appium-base-driver/blob/eba4635dcd8fe81899c3349d7e474acfb66037e7/lib/protocol/errors.js#L505

lib/css-converter.js
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/**
+ * Add android:id/ to beginning of string if it's not there already
+ *
+ * @param {string} str
+ * @returns {string} String with `android:id/` prepended (if it wasn't already)
+ */
+function prependAndroidId (str) {
+  if (!str.startsWith('android:id/')) {
```
mykola-mokhnach, line 122: return str.startsWith('android:id/') ? `android:id/${str}` : str

lib/css-converter.js
```diff
@@ -0,0 +1,307 @@
...
+ * @returns {string} The CSS selector converted to a UiSelector
+ */
+CssConverter.toUiAutomatorSelector = function toUiAutomatorSelector (cssSelector) {
+  let cssObj;
+  try {
+    cssObj = parser.parse(cssSelector);
+  } catch (e) {
+    log.errorAndThrow(`Could not parse CSS '${cssSelector}'. Reason: '${e}'`);
+  }
+  return parseCssObject(cssObj);
```
mykola-mokhnach, line 304: what if we mix these helper function into cssObj to avoid passing cssObj argument around:

```
function parseRule () {
...}
....

// was function parseCssObject (css)
function parse () {
     switch (this.type) {
     case 'rule':
       return this.parseRule();
...
}

...

cssObj.parse = parse.bind(cssObj);
cssObj.parseRule = parseRule.bind(cssObj);
```

dpgraham: I don't think we're going to agree on this one 😄  ... I'd be happy to have a third Appium contributor be a tie-breaker.

lib/css-converter.js
dpgraham: I added `throw new errors.InvalidSelectorError`

lib/css-converter.js
```diff
@@ -0,0 +1,309 @@
...
+      `Only 'string' and empty attribute types are supported. Found '${cssAttr.valueType}'`);
+  }
+  const attrName = assertGetAttrName(cssAttr);
+  const methodName = toSnakeCase(attrName);
+  if (BOOLEAN_ATTRS.includes(attrName)) {
+    return `.${methodName}(${assertGetBool(cssAttr)})`;
+  }
+
+  if (!STR_ATTRS.includes(attrName)) {
+    return;
```
mykola-mokhnach, line 150: according to the docstring this method cannot return undefined

mykola-mokhnach (approved):

lib/css-converter.js
mykola-mokhnach: I don't want to block the PR. You've done a great work and it would be a waste not to have it merged.

@jlipps Do you have any more comments or suggestions on the PR?

dpgraham: Thanks @mykola-mokhnach !

test/functional/commands/find/by-css-e2e-specs.js
jlipps: for sure, if you want to guest write an appium pro article, i'd love to feature it

jlipps (approved):
i wasn't able to give the code a very close look but i think overall approach is great! i guess we could end up doing the same for iOS too.

dpgraham:
@jlipps yeah my plan is to do this for Espresso (viewmatchers + datamatchers) and then iOS (class chain).

KazuCocoa (approved):

lib/css-converter.js
```diff
@@ -0,0 +1,319 @@
+import { CssSelectorParser } from 'css-selector-parser';
+import _ from 'lodash';
```
wswebcreation, line 2: Can't we just do this 

```suggestion
import { escapeRegExp } from 'lodash';
```

and change the `_ escapeRegExp` into `escapeRegExp` like we normally do with imports?

lib/css-converter.js
```diff
@@ -0,0 +1,319 @@
...
+ */
+function parseCssRule (cssRule) {
+  const { nestingOperator } = cssRule;
+  if (nestingOperator && nestingOperator !== ' ') {
+    throw new Error(`'${nestingOperator}' is not a supported combinator. ` +
+      `Only child combinator (>) and descendant combinator are supported.`);
+  }
+
+  let uiAutomatorSelector = 'new UiSelector()';
+  if (cssRule.tagName && cssRule.tagName !== '*') {
```
wswebcreation, line 243: Not sure how to do this differently, but I always have some issues reading multiple `if|else`-statements like this

lib/css-converter.js
```diff
@@ -0,0 +1,319 @@
...
+  if (cssAttr.valueType && cssAttr.valueType !== 'string') {
+    throw new Error(`'${cssAttr.name}=${cssAttr.value}' is an invalid attribute. ` +
+      `Only 'string' and empty attribute types are supported. Found '${cssAttr.valueType}'`);
+  }
+  const attrName = assertGetAttrName(cssAttr);
+  const methodName = toSnakeCase(attrName);
+
+  // Validate that it's a supported attribute
+  if (!STR_ATTRS.includes(attrName) && !BOOLEAN_ATTRS.includes(attrName)) {
+    throw new Error(`'${attrName}' is not a support attribute. Supported attributes are ` +
```
mykola-mokhnach, line 148: `'${attrName}' attribute is not supported`

PR #413, opened 2020-09-17
chore: android 11 PR testing

dpgraham:


mykola-mokhnach:
@dpgraham it is necessary to update the emulator startup detection script for API 30. I thought you've seen https://github.com/appium/appium-test-support/pull/42

In short, Google has changed the default `ps` command logic on the recent API versions, so it is necessary to start it with -A argument to list processes belonging to all users.
Also, for Azure the script could be modified the way it was done in https://github.com/appium/appium-adb/pull/550
There we use the fact Azure has the most recent emulator version installed, so we don't use `ps` at all, but rather just add `-delay-adb` command line option to the emulator itself (it is supported since API 28, also emulator binary itself must be up to date)

ci-jobs/scripts/start-emulator.sh
```diff
@@ -29,6 +29,11 @@ while [[ $(( `date +%s` - $secondsStarted )) -lt $TIMEOUT ]]; do
   pgrep -nf avd || exit 1
 
   processList=`adb shell ps`
+  if ! [[ $processList =~ "root " ]]; then
+        # In recent APIs running `ps` without `-A` only returns
+        # processes belonging to the current user (in this case `shell`)
+        $processList=$(adb shell ps -A)
```
mykola-mokhnach, line 35: `$` is not needed here

ci-jobs/scripts/start-emulator.sh
```diff
@@ -29,6 +29,11 @@ while [[ $(( `date +%s` - $secondsStarted )) -lt $TIMEOUT ]]; do
...
+    fi
```
mykola-mokhnach, line 36: is indentation correct?

PR #427, opened 2020-11-30
feat: add mobile command to open uiautomatorViewer

dpgraham:


lib/commands/general.js
```diff
@@ -296,6 +300,14 @@ commands.mobileType = async function mobileType (opts = {}) {
   return await this.adb.typeUnicode(text);
 };
 
+/**
+ * Opens the UiautomatorViewer
+ */
+commands.mobileUiautomatorViewer = async function mobileUiautomatorViewer () {
```
mykola-mokhnach, line 306: why such command is needed?

dpgraham: To open the uiautomatorViewer and use it to inspect the application, without the need of using the command line interface.

mykola-mokhnach: but this is an automated session. Why one would need to run the viewer during the session that does not need any manual interaction?

KazuCocoa: Agree with mykola's question.

Personally, if we add this command, this should be behind a security flag. For cloud vendors (as an example), calling this command many times could lead machine resource issue

dpgraham: Use cases:
1. Using with Appium Desktop. It runs an automated session, but if the session is left running, the inspector can be utilized.
2. Using with the Cloud Providers. I can start a session, leave it running, open the UiAutomator Viewer and inspect the Android app

I don't know if I'm concerned about performance? Seems like there's probably more effective ways to slam a VM's performance than this.

mykola-mokhnach: - Why not to implement this functionality in Appium desktop directly then?
- Cloud providers have full control over their infra. I did not see any requests for such functionality. 

Beside of that:
- The uiautomatorviewer tool itself has different paths on Windows
- IT is locked to the current Java version (e.g. it fails to start with Java11+ or if Java is not installed/found)

PR #428, opened 2020-12-16
docs: Document UiAutomator2 capabiltties

mykola-mokhnach:
I want to update README, so it contains all the necessary info that helps one to use the driver effectively. This will be the first step


jlipps (approved):
awesome!

KazuCocoa (approved):

PR #440, opened 2021-02-07
fix: Align autoLaunch capability behaviour with iOS

mykola-mokhnach:
iOS does app install, but does not launch it, which makes more sense in the capability name context.


KazuCocoa (approved):
Seems good.

https://github.com/appium/appium-uiautomator2-driver/blob/819a064acf3c5da0f29faeccb99c4772e6b0fb08/lib/driver.js#L419-L423

is responsible for launching the app under test

PR #452, opened 2021-04-16
feat: set 'disableAnalytics true'

KazuCocoa:
https://github.com/appium/appium/issues/15274
It is reasonable to disable this to prevent unexpected fatal error.

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -266,8 +266,8 @@ class UiAutomator2Server {
       cmd.push('-e', 'DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES', this.disableSuppressAccessibilityService);
     }
     cmd.push(INSTRUMENTATION_TARGET);
-    # Disable Google analytics
-    cmd.push('-e', 'disableAnalytics', true)
+    // Disable Google analytics to prevent possible fatal exception
+    cmd.push('-e', 'disableAnalytics', true);
     const instrumentationProcess = this.adb.createSubProcess(['shell', ...cmd]);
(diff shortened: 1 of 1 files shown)


lib/uiautomator2.js
```diff
@@ -266,6 +266,8 @@ class UiAutomator2Server {
       cmd.push('-e', 'DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES', this.disableSuppressAccessibilityService);
     }
     cmd.push(INSTRUMENTATION_TARGET);
+    // Disable Google analytics to prevent possible fatal exception
+    cmd.push('-e', 'disableAnalytics', true);
```
mykola-mokhnach, line 270: shouldn't we push it before setting the target?

KazuCocoa: oh, that's right

mykola-mokhnach (approved):

lib/uiautomator2.js
```diff
@@ -265,6 +265,8 @@ class UiAutomator2Server {
     if (_.isBoolean(this.disableSuppressAccessibilityService)) {
       cmd.push('-e', 'DISABLE_SUPPRESS_ACCESSIBILITY_SERVICES', this.disableSuppressAccessibilityService);
     }
+    // Disable Google analytics to prevent possible fatal exception
+    cmd.push('-e', 'disableAnalytics', true);
```
mykola-mokhnach, line 269: I assume `true` should be a string

KazuCocoa: link: https://github.com/appium/appium-espresso-driver/pull/666#discussion_r614686105

PR #454, opened 2021-05-04
chore: Remove dependency on yargs

mykola-mokhnach:


KazuCocoa (approved):
oh, this change itself looks ok, but https://github.com/appium/appium-support/pull/223 affected test code.
(am fine to merge this and the test fix will be a new PR)

mykola-mokhnach:
I have removed this integration test since it does the same that is done in e2e tests of appium-support lib. There is no point to repeat it here as well

PR #457, opened 2021-05-14
feat: Add active element support

vane:
Add support to retrieve active element from android application.
Corresponding PR.
https://github.com/appium/appium-uiautomator2-server/pull/423

### DIFF
--- a/lib/commands/element.js
+++ b/lib/commands/element.js
@@ -9,6 +9,10 @@ function toBool (s) {
   return _.isString(s) ? (s.toLowerCase() === 'true') : !!s;
 }
 
+commands.active = async function active () {
+  return await this.uiautomator2.jwproxy.command('/element/active', 'GET');
+};
+
 commands.getAttribute = async function (attribute, elementId) {
   return await this.uiautomator2.jwproxy.command(`/element/${elementId}/attribute/${attribute}`, 'GET', {});
 };
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "appium-base-driver": "^7.0.0",
     "appium-chromedriver": "^4.23.1",
     "appium-support": "^2.49.0",
-    "appium-uiautomator2-server": "^4.20.0",
+    "appium-uiautomator2-server": "^4.22.1",
     "asyncbox": "^2.3.1",
     "axios": "^0.x",
     "bluebird": "^3.5.1",
--- a/test/functional/commands/general/element-e2e-specs.js
+++ b/test/functional/commands/general/element-e2e-specs.js
@@ -32,4 +32,12 @@ describe('element', function () {
       await el.text().should.eventually.equal('original value');
     });
   });
+
+  describe('active', function () {
+    it('should active element be equal to clicked element', async function () {
+      await el.click();
+      const activeElement = await driver.active();
+      activeElement.value.should.equal(el.value);
+    });
+  });
 });


lib/commands/general.js
```diff
@@ -9,6 +9,10 @@ let extensions = {},
     commands = {},
     helpers = {};
 
+commands.active = async function () {
```
jlipps, line 12: is there an element commands file? doesn't _really_ matter where this goes but if there's an elements file that would be more sensible I think.

mykola-mokhnach:
Please bump the minimum appium-uiautomator2-server component version in package.json (should be `^4.22.0`)

lib/commands/general.js
```diff
@@ -9,6 +9,10 @@ let extensions = {},
...
+  return await this.uiautomator2.jwproxy.command('/element/active', 'POST', {});
```
mykola-mokhnach, line 13: the request method must be GET according to the W3C protocol: https://www.w3.org/TR/webdriver/#dfn-get-active-element

lib/commands/general.js
vane: Thanks for feedback. I put it to `general.js` to be consistent with `appium-ios-driver` project but I can move it to element.

lib/commands/general.js
vane: hmm there is call for active element in `appium-xcuitest-driver` that is using `POST` method I wanted to get it consistent with that method https://github.com/appium/appium-xcuitest-driver/blob/70adbf60bbf66deb86388e602b53eae902a3e04f/lib/driver.js#L100

jzaleski: Per the comment on https://github.com/appium/appium-uiautomator2-server/pull/424 we'll submit a PR to deprecate the `POST` of `/active` if it still exists on _XCUITest_ and add the `GET` variant.

lib/commands/general.js
vane: Ok I moved it to `element.js` and changed to GET.

jzaleski:
I believe this: https://github.com/appium/appium-uiautomator2-server/pull/424 will mandate another server-version bump so per @mykola-mokhnach's comment above please set the minimum-version in the package.json to whatever that is (presumably `4.22.1`).

jzaleski:
I would assume the validation checks will pass once https://github.com/appium/appium-uiautomator2-server/pull/424 is merged as right now version `4.22.1` of the _Server_ component does not exist.

lib/commands/element.js
```diff
@@ -9,6 +9,10 @@ function toBool (s) {
   return _.isString(s) ? (s.toLowerCase() === 'true') : !!s;
 }
 
+commands.active = async function () {
```
mykola-mokhnach, line 12: please add a name to the anonymous function - it makes easier to debug the the stuff:

`async function ()` -> `async function active ()`

mykola-mokhnach:
The patch to the server package with version 4.22.1 has been published

lib/commands/element.js
vane: Right, fixed. Sorry for trouble with w3c stuff.

mykola-mokhnach: np, that is why we have the review process :)

mykola-mokhnach (approved):

lib/commands/element.js
```diff
@@ -9,6 +9,10 @@ function toBool (s) {
   return _.isString(s) ? (s.toLowerCase() === 'true') : !!s;
 }
 
+commands.active = async function active () {
+  return await this.uiautomator2.jwproxy.command('/element/active', 'GET', {});
```
mykola-mokhnach, line 13: the last argument is redundant here since GET requests don't have bodies

mykola-mokhnach:
@vane Would it be possible to add a simple end-to-end test for this feature?

lib/commands/element.js
vane: removed

vane:
Added simple e2e test as requested

jlipps:
awesome work, thanks @vane !

PR #459, opened 2021-05-21
feat: add ability for driver to recieve its own cli args

khanayan123:
depends on https://github.com/appium/appium-android-driver/pull/693


lib/driver.js
```diff
@@ -135,11 +135,15 @@ const MEMOIZED_FUNCTIONS = [
 ];
 
 class AndroidUiautomator2Driver extends BaseDriver {
-  constructor (opts = {}, shouldValidateCaps = true) {
+  constructor (opts = {}, shouldValidateCaps = true, driverArgs = {}) {
     // `shell` overwrites adb.shell, so remove
     delete opts.shell;
+    opts.reboot = helpers.parseReboot(driverArgs);
```
mykola-mokhnach, line 141: one universal function like 

```js
function parseKnownArgs (allArgs, knownArgNames) {
  return allArgs.reduce((acc, argName) => {
    if (knownArgNames.includes(argName)) {
      acc[argName] = allArgs[argName];
    }
    return acc;
  }, {});
}
```

should be enough

mykola-mokhnach: then to call it: super(opts, shouldValidateCaps, Object.assign(opts, parseKnownArgs(driverArgs, ['reboot', 'chromedriverPort'...])));

mykola-mokhnach: Also, if these are just capabilities then why cannot we simply pass default capabilities map to the appropriate driver?

jlipps: @mykola-mokhnach in many cases they won't be caps. they'll be things the server admin would need to set on the CLI, that we wouldn't want accessible as caps since we don't want end users fiddling with them.

mykola-mokhnach: in such case it would still make sense to have them in a map rather in separate entries. The only diff would be that they will be applied after user capabilities rather than before

jlipps: i'm confused; the `driverArgs` object is a map of names to values.

jlipps (approved):
Looks like we'll need to run CI once the function is actually exported in basedriver.

mykola-mokhnach:
Similar comments here as for https://github.com/appium/appium-xcuitest-driver/pull/1307

mykola-mokhnach:
Closed as obsolete

PR #463, opened 2021-06-04
fix: Update properly name in session cleanup

mykola-mokhnach:
The property name is /sessions API output is `sessionId` rather than `id`


KazuCocoa (approved):
oh, i see https://github.com/appium/appium-uiautomator2-server/blob/d3cd4a3baa4209d02afcdd2fc0007717890bf138/app/src/main/java/io/appium/uiautomator2/model/api/SessionModel.java#L21

jlipps (approved):

PR #481, opened 2021-11-17
feat: add se log route as no-proxy

KazuCocoa:
for https://github.com/appium/appium/issues/16106 as UIA2

Will do the same change in the espresso driver later.


mykola-mokhnach (approved):

mykola-mokhnach:
@KazuCocoa Should a similar change be done for xcuitest driver as well?

KazuCocoa:
Not necessary, xcuitest worked for both.
It already had `/log/` and `/logs/`. `se` ones also matched each.
I'm fine to add `/se/log/` explicitly in addition to existing `/log/`, but this is not necessary.

PR #485, opened 2021-12-30
fix: Respect the default app package while building android locators from css ones

mykola-mokhnach:
Related to https://github.com/appium/java-client/issues/1621

### DIFF
--- a/lib/css-converter.js
+++ b/lib/css-converter.js
@@ -149,10 +149,10 @@ class CssConverter {
   }
 
   /**
-   * Add android:id/ to beginning of string if it's not there already
+   * Add `<pkgName>:id/` prefix to beginning of string if it's not there already
    *
    * @param {string} locator The initial locator
-   * @returns {string} String with `android:id/` prepended (if it wasn't already)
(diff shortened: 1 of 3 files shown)


lib/css-converter.js
```diff
@@ -195,126 +141,188 @@ function parseAttr (cssAttr) {
...
-    return `.${toSnakeCase(pseudoName)}(${assertGetBool(cssPseudo)})`;
+  constructor (selector, pkg) {
+    this.selector = selector;
+    this.pkg = pkg;
   }
 
-  if (NUMERIC_ATTRS.includes(pseudoName)) {
-    return `.${pseudoName}(${cssPseudo.value})`;
+  /**
+   * Add android:id/ to beginning of string if it's not there already
```
KazuCocoa, line 152: Should we address `this.pkg` case as well?

KazuCocoa (approved):

lib/css-converter.js
mykola-mokhnach: updated

PR #487, opened 2022-01-09
docs: Document enforceXPath1 setting

mykola-mokhnach:
Based on https://github.com/appium/appium-uiautomator2-server/pull/452


README.md
```diff
@@ -269,6 +269,7 @@ mjpegScalingFactor | int | The percentage value used to apply downscaling on the
...
+enforceXPath1 | boolean | Since UiAutomator2 driver version `4.25.0` XPath2 is set as the default interpreter for the corresponding element locators. This interpreter is based on [Psychopath XPath2](https://wiki.eclipse.org/PsychoPathXPathProcessor) implementation, which is now a part of the Eclipse foundation. In most of the cases XPath1 locators are also valid XPath2 locators, so there should be no issues while locating elements. Although, since the XPath2 standard is much more advanced and complicated in comparison to the previous version, some [issues](https://github.com/appium/appium/issues/16142) are possible, which cannot be fixed easily, as we depend on the third-party library mentioned above. Then try to workaround such issues by enforcing XPath1 usage (whose implementation is a part of the Android platform itself) and assigning this setting to `true`. Note, this setting is actually applied at the time when the element lookup by XPath is executed, so you could switch it on or off whenever needed throughout your automated testing session.
```
KazuCocoa, line 272: Should we address this as _not recommended as possible_?

KazuCocoa (approved):
left a minor comment

README.md
mykola-mokhnach: I don't see a need for this. If someone wants to still use XPath1 then why should we not advice that?

KazuCocoa: I thought we'll support XPath2 only in the future.
Then, addressing something such as _xpath2 will be a primary way_ or _recommended way_ could reduce confusion to update XPath 1 to XPath 2 (if needed) on users' side since this capability could think UIA2 will support both XPath 1 and 2. So users do not need to update their xpath to 2 when they get an error in XPath2 without trying to update XPath to 2.

Or probably when we drop xpath1, we need to add a deprecation message and support both for a while with this capability(?)

mykola-mokhnach: Makes sense. Will update the description

KazuCocoa: thanks

PR #498, opened 2022-01-25
fix: remove BASEDRIVER_HANDLED_SETTINGS since no longer exists in the base driver

KazuCocoa:
I noticed the BASEDRIVER_HANDLED_SETTINGS is no longer in the base driver since the settings are in images plugin. https://github.com/appium/appium/pull/16368

So, this driver no longer needs BASEDRIVER_HANDLED_SETTINGS for now, which is driver based settings.
UIA2 driver only has plugins for the uia2 server one.

For example, test failures in https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=19412&view=results (to use appium 2.0 as its test) failed by this undefined BASEDRIVER_HANDLED_SETTINGS

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import log from '../logger';
 import B from 'bluebird';
-import { errors, BASEDRIVER_HANDLED_SETTINGS } from '@appium/base-driver';
+import { errors } from '@appium/base-driver';
 import { fs, tempDir } from '@appium/support';
 import { APK_EXTENSION } from '../extensions';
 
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
AFAIK this change was done because uia2 throws an exception if an unknown setting name is supplied (which is the case while supplying image-related settings). I'm not quite sure how that is supposed to work now. @jlipps could you please provide more details?

KazuCocoa:
ah.. i see. I tested with below:

```
@driver.update_settings(
          {
            defaultImageTemplateScale: 4,
            imageMatchThreshold: 0.9,
            actionAcknowledgmentTimeout: 200
          }
        )
```

Then, only `actionAcknowledgmentTimeout` should be applied if no images plugin.
This PR returns an error. https://github.com/appium/appium-uiautomator2-driver/pull/499 did not apply all of them.

So, we need to update the server side to apply in this case

KazuCocoa:
ah, maybe this change + server-side change to apply acceptable settings only as same as xcuitest-driver...?

KazuCocoa:
mm, no. The xcuitest-driver is still v1's base driver. So anyway this change and https://github.com/appium/appium-uiautomator2-driver/pull/499 are not enough.

PR #499, opened 2022-01-25
fix: we need to just accept any setting name

jlipps:
because plugins might need them. let's not try to be too clever.

PR #502, opened 2022-02-07
fix: Update the way driver settings are updated

mykola-mokhnach:
This is basically what https://github.com/appium/appium-uiautomator2-driver/pull/499 is supposed to do, just quite simplified in order to fix the original issue ASAP.


jlipps (commented):
thanks for taking care of this.

PR #506, opened 2022-03-17
fix: Read package name from web view info if not provided explicitly by updating android driver

KazuCocoa:
So far, our each driver release includes `shrinkwrap` to specify dependencies.
https://github.com/appium/appium-uiautomator2-driver/blob/master/.github/workflows/publish.js.yml

It needs to push a new version when we'd like to include dependencies's update like https://github.com/appium/appium-android-driver/pull/732
(This PR change itself does not need, but it needs to kick a release task)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@appium/support": "^2.55.3",
     "@babel/runtime": "^7.0.0",
     "appium-adb": "^9.0.0",
-    "appium-android-driver": "^5.0.0",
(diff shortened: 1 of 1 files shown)


jlipps (approved):

jlipps:
should we set up renovate on this repo so it does things like this automatically?

KazuCocoa:
ah, sounds good. > renovate

PR #507, opened 2022-03-25
fix: Do not force stop app on session quit if noReset is true

mykola-mokhnach:


KazuCocoa (approved):

ansrl89:
Hello, @mykola-mokhnach. 

I want to send you my opinion cause I am in trouble with this change you made.

The Appium scripts I made cannot terminate test apps by `Webdriver.quit()` after this change.
My caps is like under,
{
  ...
  "noReset": True,
  "dontStopAppOnReset": False,
  "autoLaunch": True, # (default)
  ...
}

Before the change, it works like this
1. Createing A Session with launching an app
2. ... do test ...
3. Webdriver.Quit() - deleting the session with terminating the app.

After the change,
-> 3. Webdriver.Quit() - deleting the session without terminating the app.

I also know there are some methods like `terminateApp`, `closeApp` ... etc.
However, I don't think that `noReset` can effects launching and terminating.
What I think the role of `noReset` is to clean data. 
Force restarting by cleaning data is just a side effect. This does not mean `noReset` is working for launching and terminating.

I hope this change is to be reconsidered.

thanks.

Issue #542, opened 2022-09-06
create an appium 2 driver script to clean device artifacts


jlipps:
One common problem driver users encounter is old or stale Appium-related artifacts on a device, things like:

- the uia2 server
- the appium settings app
- something else i can't remember

We often encounter these issues in issue reports and provide a long list of commands to run to uninstall these artifacts. We could wrap up this logic as a driver script and document it as a troubleshooting step, saving ourselves and users time. It could look like one of these options from a user perspective, for example:

- `appium driver run uiautomator2 clean`
- `appium driver run uiautomator2 reset`
- `appium driver run uiautomator2 clear-artifacts`

Then the adb scripts would be run under the hood (perhaps via our ADB library for robustness)

mykola-mokhnach:
`appium driver run uiautomator2 reset` script works as expected

PR #544, opened 2022-09-07
feat: Add reset script

mykola-mokhnach:
https://github.com/appium/appium-uiautomator2-driver/issues/542

### DIFF
--- a/scripts/reset.js
+++ b/scripts/reset.js
@@ -1,0 +1,26 @@
+const ADB = require('appium-adb');
+const B = require('bluebird');
+const { flatMap } = require('lodash');
+
+const SERVER_PKGS = [
+  'io.appium.uiautomator2.server',
+  'io.appium.uiautomator2.server.test',
+  'io.appium.settings',
+];
+
+async function runReset () {
+  const adb = new ADB();
+  const udids = (await adb.getConnectedDevices())
+    .filter(({state}) => state === 'device')
+    .map(({udid}) => udid);
+  if (0 === udids.length) {
+    return;
+  }
+  const uninstallPromises = udids.map((udid) => {
+    adb.curDeviceId = udid;
+    return SERVER_PKGS.map((pkgId) => adb.uninstallApk(pkgId));
+  });
+  await B.all(flatMap(uninstallPromises));
+}
+
+(async () => await runReset())();
--- a/scripts/reset.js
+++ b/scripts/reset.js
@@ -20,7 +20,7 @@ async function runReset () {
   log.info(`About to perform reset for the following device${udids.length === 1 ? '' : 's'}: ${udids}`);
   const uninstallPromises = [];
   for (const udid of udids) {
-    const deviceAdb = await ADB.createADB();
+    const deviceAdb = udids.length === 1 ? adb : await ADB.createADB();
     deviceAdb.setDeviceId(udid);
     uninstallPromises.push(...(SERVER_PKGS.map((pkgId) => deviceAdb.uninstallApk(pkgId))));
   }
--- a/scripts/reset.js
+++ b/scripts/reset.js
@@ -1,6 +1,6 @@
-const ADB = require('appium-adb');
+const { ADB } = require('appium-adb');
 const B = require('bluebird');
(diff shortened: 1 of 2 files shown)


scripts/reset.js
```diff
@@ -0,0 +1,26 @@
...
+  'io.appium.settings',
+];
+
+async function runReset () {
+  const adb = new ADB();
+  const udids = (await adb.getConnectedDevices())
+    .filter(({state}) => state === 'device')
+    .map(({udid}) => udid);
+  if (0 === udids.length) {
+    return;
```
jlipps, line 17: maybe we should log a warning like `Warning: no connected devices found. Reset not run`?

scripts/reset.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  const uninstallPromises = udids.map((udid) => {
```
jlipps, line 19: maybe we should add a log line like `Running reset on device ${udid}``

scripts/reset.js
```diff
@@ -0,0 +1,26 @@
...
+    adb.curDeviceId = udid;
+    return SERVER_PKGS.map((pkgId) => adb.uninstallApk(pkgId));
+  });
+  await B.all(flatMap(uninstallPromises));
+}
+
+(async () => await runReset())();
```
jlipps, line 26: might want to do:

```js
const {asyncify} = require('asyncbox')
// ...
asyncify(runReset)
```

to make sure errors are caught/logged/thrown appropriately

jlipps: or even just

```
runReset().catch((err) => throw err)
```

mykola-mokhnach: isn't asyncify doing the same?

jlipps: it might be, i forget if it does something special like log to stderr. i usually just do `fn().catch(err => throw err)` nowadays.

mykola-mokhnach: linter is unhappy about that syntax

scripts/reset.js
mykola-mokhnach: We have adb logging the stuff for us:

```
% npm run reset

> appium-uiautomator2-driver@2.5.0 reset
> node ./scripts/reset.js

dbug ADB Getting connected devices
dbug ADB No connected devices have been detected
```

scripts/reset.js
mykola-mokhnach: Added

scripts/reset.js
jlipps: gotta put braces around `throw` i guess, just do whatever makes lint happy.

mykola-mokhnach: 🤷 
For me it is just working as is.
I don't know what else to change

jlipps: leaving as-is will create unhandled rejections.

try this:

```js
runReset().catch((err) => {  throw err; });
```

mykola-mokhnach: I've deliberately added `throw` to the function and got the following result with the current implementation:

```
/Users/elf/code/appium-uiautomator2-driver/scripts/reset.js:12
  throw new Error('yolo');
        ^
Error: yolo
    at runReset (/Users/elf/code/appium-uiautomator2-driver/scripts/reset.js:12:9)
    at /Users/elf/code/appium-uiautomator2-driver/scripts/reset.js:31:20
    at Object.<anonymous> (/Users/elf/code/appium-uiautomator2-driver/scripts/reset.js:31:31)
    at Module._compile (node:internal/modules/cjs/loader:1120:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1174:10)
    at Module.load (node:internal/modules/cjs/loader:998:32)
    at Function.Module._load (node:internal/modules/cjs/loader:839:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
```

where an unhandled exception is observed?

jlipps: ok, guess i'm wrong about that! think we're good to go then

jlipps (approved):

PR #569, opened 2022-12-07
feat: Add forceAppLaunch capability

mykola-mokhnach:
I have aligned the actual behavior with xcuitest driver there. So the logic is:
- If noReset is true and the app is not running then we always start the app. The combination of `dontStopAppOnReset` and `forceAppLaunch` caps defines whether the app should be stopped first if it is already running
- If noReset is true and the app is running then forceAppLaunch defines whether the app should be restarted. By default it is left in the current state
- If noReset is false then we always start the app. The combination of `dontStopAppOnReset` and `forceAppLaunch` caps defines whether the app should be stopped first if it is already running

Related to https://github.com/appium/appium/issues/17594


lib/driver.js
```diff
@@ -575,28 +575,29 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
+      activity: this.opts.appActivity,
+      action: this.opts.intentAction || 'android.intent.action.MAIN',
+      category: this.opts.intentCategory || 'android.intent.category.LAUNCHER',
+      flags: this.opts.intentFlags || '0x10200000', // FLAG_ACTIVITY_NEW_TASK | FLAG_ACTIVITY_RESET_TASK_IF_NEEDED
+      waitPkg: this.opts.appWaitPackage,
+      waitActivity: this.opts.appWaitActivity,
+      waitForLaunch: this.opts.appWaitForLaunch,
+      waitDuration: this.opts.appWaitDuration,
+      optionalIntentArguments: this.opts.optionalIntentArguments,
+      stopApp: this.opts.forceAppLaunch || !this.opts.dontStopAppOnReset,
```
KazuCocoa, line 597: ✅

KazuCocoa (approved):
lgtm

jlipps (commented):
nice!

PR #576, opened 2023-01-20
ci: fix and enable some commands e2e specs

rerorero:
I just started looking into https://github.com/appium/appium/issues/16402
This PR aims for spec files directly under `test/functional/commands` dir. Subdirectories are not included.
I confirmed `npm run e2e-test:commands` passed on my laptop with `android-30;google_apis;arm64-v8a` emulator.

I have a question, what is `process.env.TESTOBJECT_E2E_TESTS` for? Do you still use it?

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -68,6 +68,15 @@ jobs:
         api-level: ${{ matrix.apiLevel }}
         disable-spellchecker: true
         target: ${{ matrix.emuTag }}
+    - uses: reactivecircus/android-emulator-runner@v2
+      name: e2e_api${{ matrix.apiLevel }}
+      with:
+        script: npm run e2e-test:commands
+        avd-name: ${{ env.ANDROID_AVD }}
+        sdcard-path-or-size: 1500M
+        api-level: ${{ matrix.apiLevel }}
+        disable-spellchecker: true
+        target: ${{ matrix.emuTag }}
     - name: Save logcat output
       if: ${{ always() }}
       uses: actions/upload-artifact@master
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -6,6 +6,7 @@ on: [pull_request]
 jobs:
   test:
     strategy:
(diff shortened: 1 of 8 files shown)


KazuCocoa:
> process.env.TESTOBJECT_E2E_TESTS

It is maybe a legacy parameter recently we no longer use. TestObject is a service SauceLabs acquired long before.

rerorero:
Got it, thank you for checking. Then I'll try to clean up them as well.

.github/workflows/functional-test.yml
```diff
@@ -62,7 +64,9 @@ jobs:
     - uses: reactivecircus/android-emulator-runner@v2
       name: e2e_api${{ matrix.apiLevel }}
       with:
-        script: npm run e2e-test:driver
+        script: |
+          npm run e2e-test:driver
+          npm run e2e-test:commands
```
rerorero, line 69: Tried to have it [run in a different step ](https://github.com/appium/appium-uiautomator2-driver/pull/576/commits/d55d87e32747af34a7a00dfeee559c211ab2f31d) resulted in [a connection error to Appium server ](https://github.com/appium/appium-uiautomator2-driver/actions/runs/3968349318/jobs/6806217843)🤔 
Please let me know if you have better approaches to run multiple tests.

rerorero:
@KazuCocoa Do you mind triggering the CI?

KazuCocoa:
I've updated the actions' permission to require only a new github user. The next push won't require an approval

rerorero:
That's very helpful thank you!!

rerorero:
hum.. some tests seem to be flaky. Can you try `Re-run failed jobs`?

KazuCocoa (approved):
👍 
It's sufficient to pass a few devices. Others could be slow emulators etc. I'll merge after checking the test results

KazuCocoa:
Hm, errors are rotation and attributes things. They might be related to emulator versions etc.
We can merge this right now since at least two versions have passed completely. Better than before

KazuCocoa:
Thank you!

PR #578, opened 2023-02-01
ci: fix functional/commands/find specs

rerorero:
a part of https://github.com/appium/appium/issues/16402
this PR aims for tests under `functional/commands/find`


test/functional/commands/find/by-css-e2e-specs.js
```diff
@@ -16,67 +16,72 @@ describe('Find - CSS', function () {
...
   });
   it('should find an element by id (android resource-id)', async function () {
-    await driver.elementByCss('#text1').should.eventually.exist;
-    await driver.elementByCss('*[id="android:id/text1"]').should.eventually.exist;
-    await driver.elementByCss('*[resource-id="text1"]').should.eventually.exist;
+    // TODO: check the expected behavior when package is omitted. id in pagesrouce is 'andoird:id/text1' but
+    // the driver prepends a package name like 'io.appium.android.apis:id/text1'.
+    // await driver.$('#text1').elementId.should.eventually.exist;
+    await driver.$('*[id="android:id/text1"]').elementId.should.eventually.exist;
+    // await driver.$('*[resource-id="text1"]').elementId.should.eventually.exist;
```
rerorero, line 23: the id in the pagesource is `io.appium.android.apis:id/text1` but the driver attempts to find an element by `io.appium.android.apis:id/text1` because pkg is passed here
https://github.com/appium/appium-uiautomator2-driver/blob/935a4e89de48942bb8cb8d9794ac0154ccec080f/lib/css-converter.js#L160

Is this expected and can we just fix the test case?

test/functional/commands/find/by-css-e2e-specs.js
```diff
@@ -16,67 +16,72 @@ describe('Find - CSS', function () {
     await deleteSession();
   });
   it('should find an element by id (android resource-id)', async function () {
-    await driver.elementByCss('#text1').should.eventually.exist;
-    await driver.elementByCss('*[id="android:id/text1"]').should.eventually.exist;
-    await driver.elementByCss('*[resource-id="text1"]').should.eventually.exist;
+    // TODO: check the expected behavior when package is omitted. id in pagesrouce is 'andoird:id/text1' but
+    // the driver prepends a package name like 'io.appium.android.apis:id/text1'.
+    // await driver.$('#text1').elementId.should.eventually.exist;
```
rerorero, line 21: This line fails. Is this expected behavior or how should I fix?

test/functional/commands/find/by-css-e2e-specs.js
```diff
@@ -16,67 +16,72 @@ describe('Find - CSS', function () {
...
+    const text = await el.getText();
     text.toLowerCase().should.equal('api demos');
   });
   it('should find an element with a chain of attributes and pseudo-classes', async function () {
-    let el = await driver.elementByCss('android.widget.TextView[clickable=true]:nth-child(1)');
-    await el.text().should.eventually.equal('Accessibility');
+    // TODO: webdriver selects 'class name' strategy.
+    // ref. https://github.com/webdriverio/webdriverio/blob/eba541a77dbc42173717e1c106a7c4d3ccb198f5/packages/webdriverio/src/utils/findStrategy.ts#L91-L96
+    this.skip();
+    let el = await driver.$('android.widget.TextView[clickable=true]:nth-child(1)');
```
rerorero, line 45: webdriverio selects 'class name' strategy here. Is this a bug of webdriverio?
https://github.com/webdriverio/webdriverio/blob/eba541a77dbc42173717e1c106a7c4d3ccb198f5/packages/webdriverio/src/utils/findStrategy.ts#L91-L96

test/functional/commands/find/by-image-e2e-specs.js
```diff
@@ -13,14 +13,15 @@ const START_IMG = path.resolve(__dirname, '..', '..', 'assets', 'start-button.pn
...
 
-describe('Find - Image @skip-ci', function () {
+describe('Find - Image', function () {
   let driver;
 
   before(async function () {
-    driver = await initSession({
-      ...APIDEMOS_CAPS,
-      appActivity: '.view.ChronometerDemo'
+    this.skip();
```
rerorero, line 20: These tests failed due to `strategy "-image" is invalid` error. I wonder if UIA2 driver supports image strategy. 
https://github.com/appium/appium-uiautomator2-driver/blob/935a4e89de48942bb8cb8d9794ac0154ccec080f/lib/driver.js#L156-L161

test/functional/commands/geo-location-e2e-specs.js
```diff
@@ -12,6 +13,12 @@ chai.use(chaiAsPromised);
 describe('geo-location -', function () {
   let driver;
   beforeEach(async function () {
+    const adb = new ADB();
+    if (await adb.getApiLevel() >= 30) {
+      // TODO: always fail with API 30, is it a bug?
+      return this.skip();
```
rerorero, line 19: I realized this test always fails with API30. I can't reproduce it on my local laptop though.

test/functional/commands/find/by-css-e2e-specs.js
KazuCocoa: What about giving `android:` as well? `driver.$('#andoird:id/text1')`

Or you can set `disableIdLocatorAutocompletion` settings https://github.com/appium/appium-uiautomator2-driver/#settings-api, then maybe driver.$('#andoird:id/text1')` will work

test/functional/commands/find/by-image-e2e-specs.js
KazuCocoa: The CI needs to install https://www.npmjs.com/package/@appium/images-plugin.
The skip is sufficient. Could you leave TODO as to install the images plugin?

test/functional/commands/find/by-css-e2e-specs.js
rerorero: Sorry I realized `#text1` is valid css selector value as I confirmed css-converter can handle it.
And the reason why this test fails is because same reason as https://github.com/appium/appium-uiautomator2-driver/pull/578#discussion_r1093244242

the driver automatically prepends the package name and sends a request to uia2 server like this way
```
Proxying [POST /element] to [POST http://127.0.0.1:8200/session/3a22c73d-af92-44a0-8e66-c6b350c8320f/element] with body: {"strategy":"-android uiautomator","selector":"new UiSelector().resourceId(\"io.appium.android.apis:id/text1\")","context":"","multiple":false}
```
but the resource-id is `android:id/text1`. So I tried `#android:id/text1` but it doesn't work because CssConverter assumes it's invalid as a locator.

My question is why do both the driver and the server need to care about the package name?
driver: https://github.com/appium/appium-uiautomator2-driver/blob/c88658da1897729e7e3856a88cb9314212eb1716/lib/css-converter.js#L160
server: https://github.com/appium/appium-uiautomator2-server/blob/12866ed644c68bec53cadf6599490a0619b41938/app/src/main/java/io/appium/uiautomator2/utils/ElementLocationHelpers.java#L95

test/functional/commands/find/by-image-e2e-specs.js
rerorero: Thanks! done
https://github.com/appium/appium-uiautomator2-driver/pull/578/commits/2c8940ce62f6808d67e4834a45bdd35cc9c999d0

test/functional/commands/find/by-css-e2e-specs.js
rerorero: >  removing the driver's side would solve this confusion

ah just removing it won't save the `android:id/text1` case.

mykola-mokhnach: I think you still can provide `android:id/text1` as identifier, but the colon and the slash must be escaped: `#android\\:id\\/identifier`

rerorero: wow it works thank you!

rerorero: I fixed it https://github.com/appium/appium-uiautomator2-driver/pull/578/commits/0432006e3893cc5306405d6b6ae1fc5b777019b1
but I wonder if it's fine to change because I think the intention of the original test case was to see if the omitted package part in the locator is automatically completed 🤔

KazuCocoa (approved):
thank you!

test/functional/commands/find/by-css-e2e-specs.js
KazuCocoa: The change existed over 2 years, so maybe the test must have been updated. We did not get any issue report as well, so probably it would be good to follow current implementation (so needed to update test case before)

rerorero:
Thank you for checking ❤️

PR #581, opened 2023-02-16
chore: remove unused unicodeKeyboard as uia2

KazuCocoa:
it seems like actually the capability is no longer needed for UIA2 driver. I was able to send unicode with without them, double checked.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -163,7 +163,6 @@ class AndroidUiautomator2Driver extends BaseDriver {
     this.desiredCapConstraints = desiredCapConstraints;
     this.uiautomator2 = null;
     this.jwpProxyActive = false;
-    this.defaultIME = null;
     this.jwpProxyAvoid = NO_PROXY;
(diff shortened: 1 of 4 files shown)


lib/driver.js
```diff
@@ -647,14 +646,6 @@ class AndroidUiautomator2Driver extends BaseDriver {
...
 
-      if (this.opts.unicodeKeyboard && this.opts.resetKeyboard && this.defaultIME) {
-        this.log.debug(`Resetting IME to '${this.defaultIME}'`);
-        try {
-          await this.adb.setIME(this.defaultIME);
-        } catch (err) {
-          this.log.warn(`Unable to reset IME: ${err.message}`);
-        }
-      }
       if (this.caps.androidCoverage) {
```
mykola-mokhnach, line 649: we should also probably get rid of these caps.
I've never seen them used anywhere

mykola-mokhnach (approved):

PR #611, opened 2023-05-02
Fixed wrong default behavior explanation readme

PowerOfCreation:
The README described something which was not the default behavior as being the default behavior. Only when setting  appium:skipUnlock manually to false then the described actions are taken. I fixed this in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ appium:noSign | Set it to `true` in order to skip application signing. By defaul
 
 Capability Name | Description
 --- | ---
-appium:skipUnlock | Whether to skip the check for lock screen presence (`true`). By default UiAutomator2 driver tries to detect if the device's screen is locked before starting the test and to unlock that (which sometimes might be unstable). Note, that this operation takes some time, so it is highly recommended to set this capability to `true` and disable screen locking on devices under test.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,7 +137,7 @@ appium:noSign | Set it to `true` in order to skip application signing. By defaul
 
 Capability Name | Description
 --- | ---
-appium:skipUnlock | Whether to skip the check for lock screen presence (`true`). By default UiAutomator2 driver tries to detect if the device's screen is locked before starting the test and to unlock that (which sometimes might be unstable). Note, that this operation takes some time, so it is highly recommended to set this capability to `true` and disable screen locking on devices under test.
+appium:skipUnlock | Whether to skip the check for lock screen presence (`true`). When set to false then UiAutomator2 driver tries to detect if the device's screen is locked before starting the test and to unlock that (which sometimes might be unstable). Note, that this operation takes some time, so it is highly recommended to set this capability to `true` and disable screen locking on devices under test.
```
mykola-mokhnach, line 140: it still makes senseto describe which behaviour is the default one: `The default driver behaviour is to verify the presence of the screen lock (e.g. `false` value of the capability)`

Issue #624, opened 2023-06-12
DELETE /actions is not implemented


jlipps:
WebdriverIO sends `DELETE /actions` whenever you perform actions. Currently this results in an error:

```
[HTTP] --> DELETE /session/dc29a207-b313-4647-a3f1-b7b79dfb334a/actions
[HTTP] {}
[AndroidUiautomator2Driver@5dea (dc29a207)] Driver proxy active, passing request on via HTTP proxy
[debug] [AndroidUiautomator2Driver@5dea (dc29a207)] Matched '/session/dc29a207-b313-4647-a3f1-b7b79dfb334a/actions' to command name 'releaseActions'
[debug] [AndroidUiautomator2Driver@5dea (dc29a207)] Proxying [DELETE /session/dc29a207-b313-4647-a3f1-b7b79dfb334a/actions] to [DELETE http://127.0.0.1:8200/session/4d2423a2-0d5f-4aa7-bf4f-ad009d19c81b/actions] with no body
[AndroidUiautomator2Driver@5dea (dc29a207)] Got response with status 404: {"sessionId":null,"value":{"error":"unknown command","message":"The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource","stacktrace":"io.appium.uiautomator2.common.exceptions.UnknownCommandException: The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource\n\tat io.appium.uiautomator2.http.ServerHandler.channelRead(ServerHandler.java:75)\n\tat io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366)\n\tat io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352)\n\tat io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345)\n\tat io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)\n\tat io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandler...
[debug] [W3C] Matched W3C error code 'unknown command' to UnknownCommandError
[HTTP] <-- DELETE /session/dc29a207-b313-4647-a3f1-b7b79dfb334a/actions 404 15 ms - 3530
```

This should be implemented in the driver or the server, even if it's a no-op

PR #625, opened 2023-06-13
fix: make releaseActions to a no-op

mykola-mokhnach:
Addresses https://github.com/appium/appium-uiautomator2-driver/issues/624


KazuCocoa (approved):

mykola-mokhnach:
> do we need to add this to the no proxy list? pretty sure the error i got indicated the command was proxied to the server.

Good catch. Added to no_proxy list

jlipps (approved):

PR #631, opened 2023-06-21
refactor: Optimize server packages installation

mykola-mokhnach:
It makes sense to perform some server installation actions in parallel to save some time on session startup.
The change should also make the current code more readable


jlipps (approved):

KazuCocoa (approved):
makes sence

PR #635, opened 2023-06-30
feat: add all the types

boneskull:
- uses TS instead of babel
- adds the usual tooling (no renovate)

this depends on appium/appium#18817 and appium/appium-android-driver#819.

- all mixins were changed to be object literals.
- because of the high levels of confusion around the capability objects being built in `createSession`, I renamed some of the methods it calls and attempted to otherwise make things more clear.  `createSession` returns an object which is a _superset_ of the user-defined capabilities, since it adds a bunch of device info.  see `lib/types.ts` for the type definitions of these objects; `createSession` responds with a capabilities object of type `Uiautomator2SessionCaps`
- renamed `fillDeviceDetails` to `getDeviceDetails` and hopefully made it faster by running the async operations in parallel.
- `initUiautomator2Server` returns the `Uiautomator2Server` instance (which we consume) so we know the value is defined. otherwise it still mutates the instance.
 
classes can use the `CreateResult` type argument when extending `BaseDriver` and implementing `ExternalDriver` to resolve with a custom value; if unset, the returned capabilities will contain only props as defined in the driver's constraints (plus any "base" constraints).

PR #636, opened 2023-06-30
chore: adapt to use execute methods

boneskull:
- also upgrades chromedriver
- fixes some broken types
- re-add `source-map-support`


boneskull:
Current dependencies on/for this PR:
* master
  * **PR #635** <a href="https://app.graphite.dev/github/pr/appium/appium-uiautomator2-driver/635" target="_blank">Graphite</a> 
    * **PR #636** <a href="https://app.graphite.dev/github/pr/appium/appium-uiautomator2-driver/636" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/appium/appium-uiautomator2-driver/636?utm_source=stack-comment).

boneskull:
@jlipps Remember some time ago I mentioned that applying execute methods after the fact (as in xcuitest) was difficult because `executeMethod()` provides a tuple of arguments instead of an "options object"?  I ran in to the same problem here, of course.  But what I did instead was just... _don't call `executeMethod()`_.  This isn't ideal (due to inconsistency) but it gets us to "documentation" faster, since we can both use `executeMethodMap` _and_ avoid needing to refactor a million method signatures this way.

boneskull:
(this won't pass a build until its target branch's PR passes a build)

package.json
```diff
@@ -64,15 +64,15 @@
...
+    "appium-chromedriver": "^5.5.1",
     "appium-uiautomator2-server": "^5.12.2",
     "asyncbox": "^2.3.1",
     "axios": "^1.x",
     "bluebird": "^3.5.1",
     "css-selector-parser": "^2.2.3",
     "lodash": "^4.17.4",
     "portscanner": "^2.2.0",
-    "source-map-support": "^0.x",
+    "source-map-support": "^0.5.21",
```
mykola-mokhnach, line 75: by doing this we only allow minor version updates. Do we want that? Also, in the module readme I've seen it is not needed anymore after Node 12, so can we simply get rid of it?

lib/commands/mixins.ts
```diff
@@ -11,6 +17,7 @@ import type {
 import type {EmptyObject} from 'type-fest';
 import {AndroidUiautomator2Driver} from '../driver';
 import type * as types from './types';
+import {Uiautomator2ExecuteMethodMap} from '../execute-method-map';
```
mykola-mokhnach, line 20: is Uiautomator2ExecuteMethodMap used anywhere?

jlipps (approved):
I second @mykola-mokhnach's questions but otherwise LGTM! I think it is fine in this case to do what you did and opt out of the standard executemethod call. It's only a convenience function anyway!

lib/commands/mixins.ts
boneskull: nope

PR #640, opened 2023-08-01
docs: Update ISSUE_TEMPLATE.md

Dor-bl:
update the link to appium Issues so it will show the user the choice options rather than an empty template


mykola-mokhnach:
I think we can replace this template with a proper one, so users can also report their issues here
@jlipps wdyt?

jlipps:
sure, I think now that drivers are maintained separately, it's ok for users to report issues here.

PR #656, opened 2023-09-20
chore: tune error messages for setHiddenApiPolicy 

KazuCocoa:
Tune the error message to give more hints for users

### DIFF
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -458,7 +458,18 @@ class AndroidUiautomator2Driver
     if (apiLevel >= 28) {
       // Android P
       this.log.info('Relaxing hidden api policy');
-      await this.adb.setHiddenApiPolicy('1', !!this.opts.ignoreHiddenApiPolicyError);
+      try {
+        await this.adb.setHiddenApiPolicy('1', !!this.opts.ignoreHiddenApiPolicyError);
+      } catch (err) {
+        this.log.errorAndThrow(
+          'Hidden API policy (https://developer.android.com/guide/app-compatibility/restrictions-non-sdk-interfaces) cannot be enabled. ' +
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
ignoring this error won't solve their problems, but rather postpone them. We need to invoke hidden APis anyway, so it would make more sense to just provide such customers with instructions on where to disable such checks (or warn them regarding possible consequences if they would still prefer to ignore the error)

KazuCocoa:
basically agree. hm, then what about something this

lib/driver.js
```diff
@@ -366,7 +366,14 @@ class AndroidUiautomator2Driver extends BaseDriver {
 
     if (apiLevel >= 28) { // Android P
       this.log.info('Relaxing hidden api policy');
-      await this.adb.setHiddenApiPolicy('1', !!this.opts.ignoreHiddenApiPolicyError);
+      try {
+        await this.adb.setHiddenApiPolicy('1', !!this.opts.ignoreHiddenApiPolicyError);
+      } catch (err) {
+        this.log.errorAndThrow('Failed to disable hidden api policy. You could configure ' +
```
mykola-mokhnach, line 372: Hidden APIs cannot be enabled. This might be happening because the device under test is not configured properly. See https://github.com/appium/appium/issues/13802  for more details. You could also set the 'ignoreHiddenApiPolicyError' capability to true in order to ignore this error, which might later lead to unexpected crashes of the automation server. Original error: ...

mykola-mokhnach:
@KazuCocoa Could you please rebase the PR?

KazuCocoa:
This pr changed to based on the current master with a force push right now

mykola-mokhnach (approved):

PR #699, opened 2023-12-17
feat: let mobile: getContexts gives waitForWebviewMs

KazuCocoa:
To close https://github.com/appium/appium/issues/19251 for uia2 driver

~This PR needs to bump the package.json after https://github.com/appium/appium-android-driver/pull/860, thus tghus us currently draft~ <= ok


lib/execute-method-map.ts
```diff
@@ -370,6 +370,9 @@ export const executeMethodMap = {
 
   'mobile: getContexts': {
     command: 'mobileGetContexts',
+    params: {
+      optional: ['opts'],
```
mykola-mokhnach, line 374: I believe the name of the optional param will be `waitForWebviewMs `

README.md
```diff
@@ -928,6 +928,12 @@ The actual stdout of the downstream `am` command.
 
 Retrieves a webviews mapping based on CDP endpoints
 
+#### Arguments
+
+Name | Type | Required | Description | Example
+--- | --- | --- | --- | ---
+waitForWebviewMs | number | no | Tells UiAutomator2 driver for how long (in milliseconds) to wait for web view(s) to appear since UiAutomator2 driver v2.53.0. `0`ms by default, then UiAutomator2 driver checks the WebView(s) availability once. | 10000
```
mykola-mokhnach, line 935: I think it make sense to explain in more details why this arg might be useful

lib/execute-method-map.ts
KazuCocoa: It looks like... this `params` is not necessary? Even when I removed this `params`, the method worked, actually the command name was necessary.

README.md
```diff
@@ -926,7 +926,13 @@ The actual stdout of the downstream `am` command.
...
 ### mobile: getContexts
 
-Retrieves a webviews mapping based on CDP endpoints
+Retrieves a WebViews mapping based on CDP endpoints
+
+#### Arguments
+
+Name | Type | Required | Description | Example
+--- | --- | --- | --- | ---
+waitForWebviewMs | number | no | Tells UiAutomator2 driver for how long (in milliseconds) to wait for web view(s) to appear since UiAutomator2 driver v2.53.0. Chrome process running on the device under test could fail to create `chrome_devtools_remote`, then the chromedriver raised `failed to connect to socket 'localabstract:chrome_devtools_remote'` error to UiAutomator2 driver. It could cause no WebViews found result, althought a couple of retrial succeeds in finding WebView(s). This argument helps to keep trying to get WebView(s) up to the given time milliseconds as one command call. This issue tends to occur Chrome v115 and over so far. [issues#19251](https://github.com/appium/appium/issues/19251) has more detail logs. `0`ms by default, then UiAutomator2 driver checks the WebView(s) availability once. | 10000
```
mykola-mokhnach, line 935: ... contains more details

mykola-mokhnach: If set to `0` (the default value), then UiAutomator2 driver only checks the WebView(s) availability once

mykola-mokhnach: If a Chrome process running on the device under test fails to create a connection to devtools socket, then the chromedriver raises an error similar to `failed to connect to socket 'localabstract:chrome_devtools_remote'` in UiAutomator2 driver -> the time in both parts of the sentence must be coordinated

mykola-mokhnach: It could cause no WebViews found result, althought a couple of retrials may fix it.

mykola-mokhnach (approved):

PR #703, opened 2023-12-26
ci: reduce interval

KazuCocoa:
to reduce the amount of running time: https://github.com/appium/appium-uiautomator2-driver/actions/runs/7326322375/job/19953271200


KazuCocoa:
This did not help. Basically the frequency of instrumentation process crashes reduced, thus the amount of time increased (based on past logs). Thus I guess reducing timeout stuff such as find timeout will help to improve the performance. But this did not help

PR #724, opened 2024-01-26
fix: createADB

KazuCocoa:
https://github.com/appium/appium-android-driver/pull/908

For https://github.com/appium/appium/issues/19713


mykola-mokhnach (approved):
You can do such bumps without reviews

PR #736, opened 2024-02-08
feat: add checkVersion option for mobile:installApp

KazuCocoa:
https://github.com/appium/appium-android-driver/pull/915 for uia2.
I noticed missing args about `mobile: removeApp` in the readme, so this pr includes it as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -773,6 +773,8 @@ The call is ignored if the app is not installed.
 Name | Type | Required | Description | Example
 --- | --- | --- | --- | ---
 appId | string | yes | The identifier of the application package to be removed | `my.app.id`
+timeout | number | no | The count of milliseconds to wait until the app is terminated. 20000ms by default. | 1500, 0
+keepData | boolean | no | Set to true in order to keep the application data and cache folders after uninstall. | true
 
 #### Returned Result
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -812,6 +814,7 @@ allowTestPackages | boolean | no | Set to true in order to allow test packages i
 useSdcard | boolean | no | Set to true to install the app on sdcard instead of the device memory. false by default | true
 grantPermissions | boolean | no | Set to true in order to grant all the permissions requested in the application's manifest automatically after the installation is completed under Android 6+. The targetSdkVersion in the application manifest must be greater or equal to 23 and the Android version on the device under test must be greater or equal to Android 6 (API level 23) to grant permissions. Applications whose targetSdkVersion is lower than or equal to 22 must be reisntalled to grant permissions for Android 6+ devices. false by default | true
 replace | boolean | no | Set it to false if you don't want the application to be upgraded/reinstalled if it is already present on the device, but throw an error instead. true by default | false
+checkVersion | boolean | no | Set to true, in order to skip the application installation if the device under test has greater or equal to the application version. | true
```
mykola-mokhnach, line 817: Please also mention that by default an exception is thrown upon install if the same or an older app version is already installed

KazuCocoa: I haven't tested with an older one, but the same version was installed successfully with `checkVersion` `false` (or not given) as default as `-r`. The behavior may depend on the `replace`.

mykola-mokhnach: ok, then only mention the older one

KazuCocoa: Ensured that INSTALL_FAILED_VERSION_DOWNGRADE error occurs only for downgrade even with `-r`

KazuCocoa: btw, the same version succeeded in installing without `-r` as well

mykola-mokhnach (approved):

PR #767, opened 2024-04-22
docs: Update percent value for scroll gesture

ashwink-tech:
The percent value should less than or equal to 1


KazuCocoa (approved):
thank you

PR #769, opened 2024-04-29
chore: Update dependencies in package.json

MadeinFrance:
Update dependencies in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,8 @@
   },
   "dependencies": {
     "appium-adb": "^12.0.0",
-    "appium-android-driver": "^9.1.0",
-    "appium-chromedriver": "^5.6.28",
+    "appium-android-driver": "^9.1.1",
+    "appium-chromedriver": "^5.6.46",
     "appium-uiautomator2-server": "^7.0.1",
     "asyncbox": "^3.0.0",
     "axios": "^1.6.5",


KazuCocoa:
Do you need this? This would not help existing installation in this repo since they install available latest one basically

MadeinFrance:
Hi @KazuCocoa, the security scan detected issues with `@appium/base-driver@npm:9.5.2` see the PR [merged](https://github.com/appium/appium/pull/19927).

Sorry for the direct approach but in my project I saw that `appium-chromedriver` was used for `appium-android-driver` and since I include this repo@3.1.0 in `package.json` I thought it would be a good idea to bump the version.

Which approach should I follow? Wait for new versions?

KazuCocoa:
It depends on your aim. If you need to manage this repository itself, then you could fork or we could merge this change to package.json while it does not change existing behavior except for package.json file info.

Each release package includes shrinkwrap file in order to lock dependency versions. https://github.com/appium/appium-uiautomator2-driver/blob/master/.github/workflows/publish.js.yml . You could check it via `npm pack appium-uiautomator2-driver` or in installed uia2 driver package on.you local. If you need to update such dependencies, using newer UIA2 driver is a general solution. If you need a newer release, since the shrinkwrap file is generated in a release script, we could release a new version with no changes in this repository itself.

PR #771, opened 2024-05-02
ci: use GH Action for detecting LTS Node

eglitise:
Same as https://github.com/appium/appium/pull/20059.
Also adjusted unit tests to be skipped for Markdown file changes.

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -60,7 +60,7 @@ jobs:
     - uses: actions/setup-java@v3
       with:
         distribution: 'temurin'
-        java-version: '17'
+        java-version: '11'
     - run: |
         cwd=$(pwd)
         pushd "$cwd"
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -60,7 +60,7 @@ jobs:
     - uses: actions/setup-java@v3
       with:
         distribution: 'temurin'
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (approved):

mykola-mokhnach:
It looks like functional tests don't work anymore. Could you please try also bumping the java version there to 17?

eglitise:
Seems like the architecture also needs to be changed? Unsure...

mykola-mokhnach:
>   PANIC: Avd's CPU Architecture 'x86' is not supported by the QEMU2 emulator on aarch64 host.

Yes, looks like GH switched to use ARM runners now. This should be good news through - they should perform much better than intel ones

mykola-mokhnach:
Or maybe we can switch to KVM runners: https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#running-hardware-accelerated-emulators-on-linux-runners

mykola-mokhnach:
Actually we may try to copy the related changes from https://github.com/appium/appium-uiautomator2-driver/actions/runs/8919253215/job/24495129792?pr=770

eglitise:
In that case maybe it's better if I revert the functional test changes in this PR, so they can be done entirely in Kazu's PR? Same for the Espresso driver PR.

KazuCocoa (approved):

PR #796, opened 2024-06-21
fix: uninstall installed uia2 servers if they were greater than on-going session

KazuCocoa:
https://github.com/appium/appium/issues/19799#issuecomment-2181932720

The current UIA2 driver does not uninstall UIA2 servers on the device if they are newer than newly installed servers by the UIA2. It could cause old UIA2 driver with newer UIA2 servers on the device, which is a mismatch condition. It could cause unexpected behavior.
 
Maybe... `this.adb.APP_INSTALL_STATE.NOT_INSTALLED` condition should be `this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED` to "uninstall" installed UIA2 servers?

It looks like `sinon` mock's `expects`, `never`, once etc might not be so reliable. Some could be not good. To improve condition check as test code, I have extracted condition check into two methods and added tests for them. At least `shouldUninstallServerPackages` and `canInstallServerPackages` tests should be reliable.

### DIFF
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -103,7 +103,11 @@ class UiAutomator2Server {
    * @property {string} appPath
    * @property {string} appId
    */
+
   /**
+   * If the given package information includes servers should uninstall befoer installing new UIA2 servers.
+   * For example, if the servers on the device is newer than servers UIA2 driver wants to use for the session.
+   *
    * @param {PackageInfo[]} packagesInfo
    * @returns {boolean}
    */
@@ -113,6 +117,8 @@ class UiAutomator2Server {
   }
 
   /**
+   * If the given package information includes servers which can install with current UIA2 driver has.
+   *
    * @param {PackageInfo[]} packagesInfo
    * @returns {boolean}
    */
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -105,15 +105,18 @@ class UiAutomator2Server {
    */
 
   /**
-   * If the given package information includes servers should uninstall befoer installing new UIA2 servers.
-   * For example, if one of servers on the device was newer than servers current UIA2 driver wants to
+   * Checks if server components must be installed from the device under test
+   * in scope of the current driver session.
+   *
+   * For example, if one of servers on the device under test was newer than servers current UIA2 driver wants to
    * use for the session, the UIA2 driver should uninstall the installed ones in order to avoid
-   * version mismatch between the UIA2 drier and servers on the device.
-   * Also, if the device has missing servers, current UIA2 driver should uninstall all
+   * version mismatch between the UIA2 drier and servers on the device under test.
+   * Also, if the device under test has missing servers, current UIA2 driver should uninstall all
    * servers once in order to install proper servers freshly.
    *
    * @param {PackageInfo[]} packagesInfo
-   * @returns {boolean}
+   * @returns {boolean} if any of components is already installed and the other is not installed
+   *                    or the installed one has a newer version.
    */
   shouldUninstallServerPackages(packagesInfo = []) {
     const isAnyComponentInstalled = packagesInfo.some(
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -115,7 +115,7 @@ class UiAutomator2Server {
    * servers once in order to install proper servers freshly.
    *
    * @param {PackageInfo[]} packagesInfo
-   * @returns {boolean} if any of components is already installed and the other is not installed
+   * @returns {boolean} true if any of components is already installed and the other is not installed
    *                    or the installed one has a newer version.
    */
   shouldUninstallServerPackages(packagesInfo = []) {
@@ -132,7 +132,7 @@ class UiAutomator2Server {
    * Checks if server components should be installed on the device under test in scope of the current driver session.
    *
    * @param {PackageInfo[]} packagesInfo
-   * @returns {boolean} if any of components is not installed or older than currently installed in order to
+   * @returns {boolean} true if any of components is not installed or older than currently installed in order to
    *                    install or upgrade the servers on the device under test.
    */
   shouldInstall(packagesInfo = []) {
--- a/lib/uiautomator2.js
+++ b/lib/uiautomator2.js
@@ -106,14 +106,23 @@ class UiAutomator2Server {
 
   /**
    * If the given package information includes servers should uninstall befoer installing new UIA2 servers.
-   * For example, if the servers on the device is newer than servers UIA2 driver wants to use for the session.
+   * For example, if one of servers on the device was newer than servers current UIA2 driver wants to
+   * use for the session, the UIA2 driver should uninstall the installed ones in order to avoid
+   * version mismatch between the UIA2 drier and servers on the device.
+   * Also, if the device has missing servers, current UIA2 driver should uninstall all
+   * servers once in order to install proper servers freshly.
    *
    * @param {PackageInfo[]} packagesInfo
    * @returns {boolean}
    */
   shouldUninstallServerPackages(packagesInfo = []) {
-    return (packagesInfo.some(({installState}) => installState === this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED)
-    && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
+    const isAnyComponentInstalled = packagesInfo.some(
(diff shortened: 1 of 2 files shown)


lib/uiautomator2.js
```diff
@@ -97,6 +97,32 @@ class UiAutomator2Server {
...
+   * @property {string} installState
+   * @property {string} appPath
+   * @property {string} appId
+   */
+  /**
+   * @param {PackageInfo[]} packagesInfo
+   * @returns {boolean}
+   */
+  shouldUninstallServerPackages(packagesInfo = []) {
+    return (packagesInfo.some(({installState}) => installState === this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED)
```
mykola-mokhnach, line 115: I am not sure about this change.

This is the situation we were checking before with this condition:

- if one of server components is installed and the other is not then we want to uninstall the existing one before installing both

If any of the components has a newer version installed then we don't need to remove it unless the current uia2 driver is not the newest one

mykola-mokhnach: I have read the PR description and confirmed the case addresses my last guess above.
In such case and also in order to preserve the legacy behavior the uninstall check should be changed to

```js
return (packagesInfo.some(({installState}) => installState !== this.adb.APP_INSTALL_STATE.NOT_INSTALLED)
    && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
```

mykola-mokhnach: also the shouldUninstallServerPackages  method itself should have a proper doctrings explaining its purpose since this condition is not very readable

lib/uiautomator2.js
```diff
@@ -97,6 +97,38 @@ class UiAutomator2Server {
...
+    && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
+  }
+
+  /**
+   * If the given package information includes servers which can install with current UIA2 driver has.
+   *
+   * @param {PackageInfo[]} packagesInfo
+   * @returns {boolean}
+   */
+  canInstallServerPackages (packagesInfo = []) {
```
mykola-mokhnach, line 125: it's rather shouldInstall

lib/uiautomator2.js
```diff
@@ -97,6 +97,38 @@ class UiAutomator2Server {
...
+   * @param {PackageInfo[]} packagesInfo
+   * @returns {boolean}
+   */
+  shouldUninstallServerPackages(packagesInfo = []) {
+    return (packagesInfo.some(({installState}) => installState === this.adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED)
+    && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
+  }
+
+  /**
+   * If the given package information includes servers which can install with current UIA2 driver has.
```
mykola-mokhnach, line 120: Checks if server components should be installed on the device under test in scope of the current driver session

lib/uiautomator2.js
```diff
@@ -97,6 +97,38 @@ class UiAutomator2Server {
...
 
+  /**
+   * @typedef {Object} PackageInfo
+   * @property {string} installState
+   * @property {string} appPath
+   * @property {string} appId
+   */
+
+  /**
+   * If the given package information includes servers should uninstall befoer installing new UIA2 servers.
```
mykola-mokhnach, line 108: Checks if server components must be installed from the device under test in scope of the current driver session.

...
returns {boolean} true if any of components is already installed and the other is not installed or the installed one has a newer version

lib/uiautomator2.js
```diff
@@ -97,6 +97,51 @@ class UiAutomator2Server {
...
+  }
+
+  /**
+   * Checks if server components should be installed on the device under test in scope of the current driver session.
+   *
+   * @param {PackageInfo[]} packagesInfo
+   * @returns {boolean} true if any of components is not installed or older than currently installed in order to
+   *                    install or upgrade the servers on the device under test.
+   */
+  shouldInstall(packagesInfo = []) {
```
mykola-mokhnach, line 138: shouldInstallServerPackages

lib/uiautomator2.js
```diff
@@ -116,15 +161,12 @@ class UiAutomator2Server {
     );
 
     this.log.debug(`Server packages status: ${JSON.stringify(packagesInfo)}`);
-    // Enforce server packages reinstall if any of the packages is not installed, while the other is
-    const shouldUninstallServerPackages = (packagesInfo.some(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED)
-          && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
+    // To ensure if the UIA2 driver should uninstall UI2 server packages from the device
+    // to use peoper UIA2 server versions current UIA2 server wants to use.
```
mykola-mokhnach, line 165: typo: proper

test/unit/uiautomator2-specs.js
```diff
@@ -13,6 +13,182 @@ chai.use(chaiAsPromised);
...
+      'appId': 'io.appium.uiautomator2.server.test'
+    };
+    beforeEach(function () {
+      uiautomator2 = new UiAutomator2Server(log, {
+        adb,
+        tmpDir: 'tmp',
+        systemPort: 4724,
+        host: 'localhost',
+        devicePort: 6790,
+        disableWindowAnimation: false,
```
mykola-mokhnach, line 34: do we need these specific capabilities for unit tests? they are not used anyway

test/unit/uiautomator2-specs.js
```diff
@@ -13,6 +13,182 @@ chai.use(chaiAsPromised);
...
+          'installState': adb.APP_INSTALL_STATE.OLDER_VERSION_INSTALLED,
+          ...serverTestApk
+        }
+      ]).should.eql(false);
+    }),
+    it('with no server are installed', function () {
+      uiautomator2.shouldUninstallServerPackages([
+        {
+          'installState': adb.APP_INSTALL_STATE.NOT_INSTALLED,
+          'appPath': 'path/to/appium-uiautomator2-server.apk',
```
mykola-mokhnach, line 92: I assume these items should also be replaced by constants like it was done above

test/unit/uiautomator2-specs.js
```diff
@@ -13,6 +13,182 @@ chai.use(chaiAsPromised);
...
+          'installState': adb.APP_INSTALL_STATE.OLDER_VERSION_INSTALLED,
+          ...serverTestApk
+        }
+      ]).should.eql(true);
+    }),
+    it('with no server are installed', function () {
+      uiautomator2.shouldInstall([
+        {
+          'installState': adb.APP_INSTALL_STATE.NOT_INSTALLED,
+          'appPath': 'path/to/appium-uiautomator2-server.apk',
```
mykola-mokhnach, line 180: same here

test/unit/uiautomator2-specs.js
```diff
@@ -13,6 +13,182 @@ chai.use(chaiAsPromised);
...
+      uiautomator2.shouldInstall([
+        {
+          'installState': adb.APP_INSTALL_STATE.NEWER_VERSION_INSTALLED,
+          ...serverApk
+        },
+        {
+          'installState': adb.APP_INSTALL_STATE.OLDER_VERSION_INSTALLED,
+          ...serverTestApk
+        }
+      ]).should.eql(true);
```
mykola-mokhnach, line 149: I usually prefer shoud.be.true or should.be.false as more readable option

KazuCocoa: https://github.com/appium/appium-uiautomator2-driver/pull/796/commits/8339010db2dd38eb38fe7d83f3d972ec0691d8b9

lib/uiautomator2.js
```diff
@@ -116,15 +161,12 @@ class UiAutomator2Server {
     );
 
     this.log.debug(`Server packages status: ${JSON.stringify(packagesInfo)}`);
-    // Enforce server packages reinstall if any of the packages is not installed, while the other is
-    const shouldUninstallServerPackages = (packagesInfo.some(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED)
-          && !packagesInfo.every(({installState}) => installState === this.adb.APP_INSTALL_STATE.NOT_INSTALLED));
+    // To ensure if the UIA2 driver should uninstall UI2 server packages from the device
```
mykola-mokhnach, line 164: This comment is now not really needed since we have the method itself properly documented

PR #828, opened 2024-09-17
feat: Bump appium-chromedriver

mykola-mokhnach:
This disables automatic chromedriver download upon module installation

### DIFF
--- a/.github/workflows/functional-test.yml
+++ b/.github/workflows/functional-test.yml
@@ -18,20 +18,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - chromedriverVersion: "113.0.5672.63"
-          apiLevel: 32
+        - apiLevel: 32
           emuTag: google_apis
           arch: x86_64
-        - chromedriverVersion: "83.0.4103.39"
-          apiLevel: 30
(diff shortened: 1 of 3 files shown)


.github/workflows/functional-test.yml
```diff
@@ -93,14 +89,15 @@ jobs:
...
-        CHROMEDRIVER_VERSION="${{ matrix.chromedriverVersion }}" appium driver install --source=local "$cwd"
+        appium driver install --source=local "$cwd"
         appium driver doctor uiautomator2
         nohup appium server \
           --port=$APPIUM_TEST_SERVER_PORT \
           --address=$APPIUM_TEST_SERVER_HOST \
           --relaxed-security \
           --log-no-colors \
           --log-timestamp \
+          --allow-insecure chromedriver_autodownload \
```
KazuCocoa, line 100: I think this is not necessary since we already set `--relaxed-security ` which includes all security flags stuff

KazuCocoa (approved):

.github/workflows/functional-test.yml
mykola-mokhnach: yes, makes sense

PR #873, opened 2025-03-18
fix: Explicitly wait for the server shutdown before killing it forcefully

mykola-mokhnach:


lib/uiautomator2.js
```diff
@@ -438,11 +426,35 @@ class UiAutomator2Server {
         await this.adb.killProcessesByName('uiautomator');
       } catch {}
     }
-    await waitStop();
```
KazuCocoa, line 441: What I could remember is this random process end delay could occur after the `adb.forceStop` also because according to the device's adb log, the process kill related log itself delayed after sending a kill command. I still think we need to ensure the uia2 server has ended in `cleanupAutomationLeftovers` as the method's purpose.

KazuCocoa: Let me search my past logs if the `this.adb.processExists` was reliable on such env. Maybe I have some research comment somewhere.

KazuCocoa: I couldn't find a good log if `this.adb.processExists` is reliable, but I think it should be ok.

So far, I observed the process kill delay in adb logcat before, so we need to assume `this.adb.forceStop` itself could have delay to kill the process on the OS internal level. But I assume `this.adb.processExists` itself should be ok instead of `/status` endpoint check

mykola-mokhnach: thanks for checking it.

Basically, this is what I figured out so far by analyzing the server code:

after receiving the shutdown request (which is sent as part of the DELETE session request) the server will asynchronously:
- shutdown http listener and other side servers, like mjpeg broadcaster
- exit the main Junit runner entry point
- exit the server process, which would also stop the instrumentation monitoring

The previous check (GET /status) was only verifying the first stage completion, which we actually need to wait for the last one. On my test phone the time difference between the first and the last stage completion is about 400ms.

KazuCocoa: Yea, this PR idea looks good. Let me see HS's device pool, which could be good for this testing purpose. The original issue was Samsung Android 13+, so probably I could find a device

KazuCocoa (approved):
lg

I couldn't find the original issue's device but did 100+ loop to start/terminate appium sessions with multiple Samsung Android13+ devices. They didn't have any weird delay issue, so probably good with this

PR #876, opened 2025-03-31
feat: proxy /window/new endpoint to chromedriver

KazuCocoa:
instead of https://github.com/appium/appium-android-driver/pull/991

### DIFF
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -218,11 +218,11 @@ const NO_PROXY: RouteMatcher[] = [
   // For Selenium v4 (W3C does not have this route)
   ['GET', new RegExp('^/session/[^/]+/se/log/types')],
   ['GET', new RegExp('^/session/[^/]+/window/rect')],
-  ['POST', new RegExp('^/session/[^/]+/window/new')],
   ['POST', new RegExp('^/session/[^/]+/execute/async')],
   ['POST', new RegExp('^/session/[^/]+/execute/sync')],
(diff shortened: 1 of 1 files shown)


lib/driver.ts
```diff
@@ -222,6 +222,7 @@ const NO_PROXY: RouteMatcher[] = [
   ['POST', new RegExp('^/session/[^/]+/execute/sync')],
   // For Selenium v4 (W3C does not have this route)
   ['POST', new RegExp('^/session/[^/]+/se/log')],
+  ['POST', new RegExp('^/session/[^/]+/window/new')],
```
mykola-mokhnach, line 225: not sure how this helps
if an endpoint is present in this list then the driver would execute the corresponding matching method defined for it instead of proxying by default. 
I would assume that currently the new window request does not work because we already have a regexp item in this list, which erroneously includes the `/window/new`, and which we need to tune to not include this suffix
I could be wrong though

KazuCocoa: Hm, yea. I read this naming again as `NO_PROXY`, but without this, this endpoint didn't proxy to chromedriver in webview context. Initially, I expected this endpoint was proxied to chromedriver in webview context without this as not in `CHROME_NO_PROXY`. (so I confused) Let me track again

KazuCocoa: My test was uia2 version related things. so old one had this issue but some newer one worked as expected. So, latest works as our expected (no modifications are required)

PR #911, opened 2025-07-20
fix: do not proxy commands for retrieving session data

eglitise:
This is a fix for several commands added in base-driver v9.16 (`/appium/commands`, `/appium/capabilities` and `/appium/extensions`). Seems like the driver was attempting to proxy them to the UiAutomator2 server, which is not needed.


KazuCocoa (approved):

KazuCocoa:
then, maybe espresso driver also needs the same thing

PR #914, opened 2025-08-17
chore: remove no reference dev deps rimraf

KazuCocoa:
```
kazu $ git grep 'rimraf'
package.json:    "rimraf": "^5.0.0",
```

instead of https://github.com/appium/appium-uiautomator2-driver/pull/805


KazuCocoa:
I'll merge this after CI passes as just for dev deps

PR #951, opened 2025-12-02
fix: Update mobile: screenshots endpoint to properly handle virtual display ids

mykola-mokhnach:


lib/commands/screenshot.ts
```diff
@@ -0,0 +1,146 @@
+import _ from 'lodash';
+import B from 'bluebird';
+import {imageUtil} from 'appium/support';
+import type {AndroidUiautomator2Driver} from '../driver';
+import type {Screenshot} from './types';
+import type {StringRecord} from '@appium/types';
+
+// Matches SurfaceFlinger output format:
+// Physical: Display 4619827259835644672 (HWC display 0): port=0 pnpId=GGL displayName="EMU_display_0"
+// Virtual: Display 11529215049243506835 (Virtual display): displayName="Emulator 2D Display" uniqueId="..."
```
KazuCocoa: What command did you get this `Virtual: Display 11529215049243506835...` format?

My local env didn't print this format for virtual display with `adb shell dumpsys SurfaceFlinger --display-id`, `adb shell dumpsys SurfaceFlinger --displays` or `adb shell dumpsys SurfaceFlinger`.

KazuCocoa: Gotcha. Looks like the output depends on API Level.

API Level 35 -> `--display-id` prints only physical displays. `--displays` has virtual one (as same as current uia2 server)
API Level 32 -> `--display-id` prints only physical displays. `--displays` doesn't work so we only could get virtal display only with `adb shell dumpsys SurfaceFlinger`

PR #996, opened 2026-05-03
feat: Ditch bluebird

mykola-mokhnach:


test/functional/commands/orientation-e2e-specs.ts
```diff
@@ -55,24 +73,24 @@ describe('apidemo - orientation -', function () {
     });
     it('should rotate screen to landscape', async function () {
       await driver.setOrientation('PORTRAIT');
-      await B.delay(3000);
+      await waitForOrientation('PORTRAIT');
       await driver.setOrientation('LANDSCAPE');
-      await B.delay(3000);
-      await expect(driver.getOrientation()).to.eventually.become('LANDSCAPE');
+      await waitForOrientation('LANDSCAPE');
+      expect(await driver.getOrientation()).to.eql('LANDSCAPE');
```
eglitise: Since `waitForOrientation` now calls `driver.getOrientation()`, the second line doesn't check anything new anymore. Same applies to the following two tests as well.

mykola-mokhnach: makes sense. Removed

eglitise: Now the tests don't have any expect calls, so they're not actually validating anything

mykola-mokhnach: they still check if the actual orientation has actually changed

eglitise (approved):

PR #1001, opened 2026-05-13
feat: Ditch lodash

mykola-mokhnach:


lib/commands/keyboard.ts
```diff
@@ -81,7 +80,7 @@ export async function mobileType(
   this: AndroidUiautomator2Driver,
   text: string | number | boolean,
 ): Promise<boolean> {
-  if (_.isUndefined(text)) {
+  if (text === undefined) {
     throw new errors.InvalidArgumentError(`The 'text' argument is mandatory`);
   }
```
KazuCocoa: i might forget, but the `text` is required in https://github.com/appium/appium-uiautomator2-driver/blob/f17216c6fd81115c6eef6f2f263b00ac53326426/lib/execute-method-map.ts#L120-L124, so the throw might not be necessary?

KazuCocoa (approved):

lib/commands/keyboard.ts
mykola-mokhnach: makes sense, fixed

PR #1031, opened 2026-07-29
chore: Drop chai

mykola-mokhnach:


test/functional/commands/find/by-accessibility-id-e2e-specs.ts
```diff
@@ -19,14 +16,14 @@ describe('Find - accessibility ID', function () {
     await deleteSession();
   });
   it('should find an element by name', async function () {
-    await expect(driver.$('~Animation').elementId).to.eventually.exist;
+    assert.ok((await driver.$('~Animation').elementId) != null);
```
eglitise: This and other lines use `assert.ok(elem.elemId != null)`, is `assert.ok(elem.elemId)` not sufficient?

test/functional/commands/find/by-uiautomator-e2e-specs.ts
```diff
@@ -22,85 +19,85 @@ describe('Find - uiautomator', function () {
   });
   it('should find elements with a boolean argument', async function () {
     const els = await driver.$$('android=new UiSelector().clickable(true)');
-    expect(els).to.have.length.at.least(8);
+    assert.ok((await els.length) >= 8);
```
eglitise, line 22: Is the extra `await` needed here (and for other similar lines)?

test/functional/commands/general/attribute-e2e-specs.ts
```diff
@@ -22,31 +19,31 @@ describe('apidemo - attributes', function () {
...
-    await expect(animationEl.getAttribute('contentDescription')).to.eventually.become('Animation');
+    assert.deepStrictEqual(await animationEl.getAttribute('contentDescription'), 'Animation');
   });
   it('should be able to find displayed attribute', async function () {
-    await expect(animationEl.getAttribute('displayed')).to.eventually.become('true');
+    assert.deepStrictEqual(await animationEl.getAttribute('displayed'), 'true');
   });
   it('should be able to find enabled attribute', async function () {
-    await expect(animationEl.getAttribute('enabled')).to.eventually.become('true');
+    assert.deepStrictEqual(await animationEl.getAttribute('enabled'), 'true');
```
eglitise: These can be just `strictEqual`

test/functional/commands/general/context-e2e-specs.ts
```diff
@@ -70,7 +67,7 @@ describe('apidemo - context', function () {
       });
       driver = await initSession(caps);
       const context = await driver.getContext();
-      expect(context).to.not.eql('NATIVE_APP');
+      assert.notDeepStrictEqual(context, 'NATIVE_APP');
```
eglitise: This can be just `notStrictEqual`

test/functional/commands/keyboard/keyboard-e2e-specs.ts
```diff
@@ -326,8 +323,8 @@ describe('keyboard', function () {
       // make sure the IME has been restored
       if (adb) {
         const ime = await adb.defaultIME();
-        expect(ime).to.eql(initialIME);
-        expect(ime).to.not.eql('io.appium.settings/.UnicodeIME');
+        assert.deepStrictEqual(ime, initialIME);
+        assert.notDeepStrictEqual(ime, 'io.appium.settings/.UnicodeIME');
```
eglitise: These also don't need to use `deep` checks

test/functional/commands/orientation-e2e-specs.ts
```diff
@@ -35,15 +32,15 @@ describe('apidemo - orientation -', function () {
...
+      assert.deepStrictEqual(await driver.getOrientation(), 'LANDSCAPE');
     });
     it('should have portrait orientation if nothing requested', async function () {
       driver = await initSession(
         amendCapabilities(APIDEMOS_CAPS, {
           'appium:appActivity': '.view.TextFields',
         }),
       );
-      await expect(driver.getOrientation()).to.eventually.eql('PORTRAIT');
+      assert.deepStrictEqual(await driver.getOrientation(), 'PORTRAIT');
```
eglitise: Same here

test/functional/commands/viewport-e2e-specs.ts
```diff
@@ -69,9 +66,9 @@ describe('testViewportCommands', {skip: isCi()}, function () {
     const viewB64 = Buffer.from(viewScreen as string, 'base64');
     const fullImgMeta = await sharp(fullB64).metadata();
     const viewImgMeta = await sharp(viewB64).metadata();
-    expect(viewportRect.top).to.eql(statBarHeight);
-    expect(viewImgMeta.height).to.eql(viewportRect.height);
-    expect(viewImgMeta.width).to.eql(fullImgMeta.width);
-    expect(fullImgMeta.height).to.be.above(viewImgMeta.height!);
+    assert.deepStrictEqual(viewportRect.top, statBarHeight);
+    assert.deepStrictEqual(viewImgMeta.height, viewportRect.height);
+    assert.deepStrictEqual(viewImgMeta.width, fullImgMeta.width);
```
eglitise: Same here

test/functional/commands/find/by-accessibility-id-e2e-specs.ts
mykola-mokhnach: Good catch, simplified — pushed as `assert.ok(elem.elementId)` (and the same for every other elementId/driver existence check across the diff).

test/functional/commands/find/by-uiautomator-e2e-specs.ts
mykola-mokhnach: It is, unfortunately. webdriverio types `ChainablePromiseArray.length` as `Promise<number>` (it's designed to also support chaining before the array itself is awaited), and that stays true even after `els` itself has already been awaited — TS doesn't narrow it to a plain array because `ChainablePromiseArray`'s declared interface has no `then`, so `Awaited<>` leaves it untouched. `assert.ok(els.length >= 8)` fails to compile as a result. Chai never caught this because `.to.have.length()` doesn't type-check its target at all. A few lines below in this same file the original chai-based code already worked around the same issue explicitly (`const clickableLength = await clickableEls.length;`), so this isn't a new pattern for the file, just applied more consistently now that assert forces the type to be checked. At runtime it's a no-op (`await` on an already-resolved number just returns it), so no behavior change.

test/functional/commands/general/attribute-e2e-specs.ts
mykola-mokhnach: Agreed, fixed — switched to `strictEqual` here and for every other primitive (string/number) comparison across the diff that had been mapped to `deepStrictEqual`. Left `deepStrictEqual` in place where the target is genuinely an array/object/Buffer.

test/functional/commands/general/context-e2e-specs.ts
mykola-mokhnach: Already fixed in the same push — changed to `notStrictEqual`.

test/functional/commands/keyboard/keyboard-e2e-specs.ts
mykola-mokhnach: Already fixed in the same push — both changed to `strictEqual`/`notStrictEqual`.

test/functional/commands/orientation-e2e-specs.ts
mykola-mokhnach: Already fixed in the same push — changed to `strictEqual`.

test/functional/commands/viewport-e2e-specs.ts
mykola-mokhnach: Already fixed in the same push — all three changed to `strictEqual`.

test/functional/commands/find/by-uiautomator-e2e-specs.ts
eglitise: But assertions with exact length value checks do not use `await` (e.g. `assert.strictEqual(els.length, 2)`) - is that okay?

mykola-mokhnach: Good point — that was inconsistent (it only "worked" because `strictEqual`'s first param is typed `unknown`, so it doesn't force the `Promise<number>` type to be checked the way an operator comparison does). Pushed a follow-up making every `.length` read on these chainable arrays consistently `await`ed, exact-value checks included.

test/functional/commands/general/source-e2e-specs.ts
```diff
@@ -22,13 +19,13 @@ describe('apidemo - source', function () {
   });
 
   function assertSource(source: string): void {
-    expect(source).to.exist;
+    assert.ok(source != null);
```
eglitise, line 22: Perhaps this and line 28 can also omit the null comparison

mykola-mokhnach: I would prefer to keep the current implementation. It reflects `exists` semantic better

test/unit/commands/general-specs.ts
```diff
@@ -81,13 +78,14 @@ describe('General', function () {
     });
 
     it('should reject if no apks were given', async function () {
-      await expect(driver.execute('mobile: installMultipleApks', {apks: []})).to.be.rejectedWith(
-        'No apks are given to install',
-      );
+      await assert.rejects(driver.execute('mobile: installMultipleApks', {apks: []}), (err: Error) => {
+        assert.ok(err.message.includes('No apks are given to install'));
+        return true;
```
eglitise: Can this be simplified to the regex matcher approach used in the previous test and elsewhere?

test/unit/commands/screenshot-specs.ts
```diff
@@ -10,36 +9,57 @@ describe('Screenshot - parseSurfaceFlingerDisplays', function () {
...
+          id: result['4619827259835644672'].id,
+          isDefault: result['4619827259835644672'].isDefault,
+          name: result['4619827259835644672'].name,
+        },
+        {
+          id: '4619827259835644672',
+          isDefault: true,
+          name: 'EMU_display_0',
+        },
+      );
```
eglitise: This works but I feel like a more compact (and better debuggable) approach would be to just have an equality check for each parameter:
```js
assert.strictEqual(result['4619827259835644672'].id, '4619827259835644672')
...
```
 Same for other such tests

test/unit/driver-specs.ts
```diff
@@ -58,7 +55,11 @@ describe('driver.js', function () {
             },
           } as any,
         ),
-      ).to.be.rejectedWith('does not exist or is not accessible');
+        (err: Error) => {
+          assert.ok(err.message.includes('does not exist or is not accessible'));
+          return true;
```
eglitise: This can probably also use the regex matcher form

test/unit/driver-specs.ts
```diff
@@ -142,7 +143,10 @@ describe('driver.js', function () {
       );
 
       checkAppPresentStub.restore();
-      await expect(driver.checkAppPresent()).to.be.rejectedWith('Could not find');
+      await assert.rejects(driver.checkAppPresent(), (err: Error) => {
+        assert.ok(err.message.includes('Could not find'));
+        return true;
```
eglitise: Same here

test/unit/driver-specs.ts
```diff
@@ -155,31 +159,26 @@ describe('driver.js', function () {
     });
     describe('#proxyActive', function () {
       it('should exist', function () {
-        expect(driver.proxyActive).to.be.an.instanceof(Function);
+        assert.ok(driver.proxyActive instanceof Function);
       });
       it('should return true', function () {
-        expect(driver.proxyActive('abc')).to.be.true;
-      });
-      it('should throw an error if session id is wrong', function () {
```
eglitise, line 163: Was this test intentionally removed?

test/unit/driver-specs.ts
```diff
@@ -269,23 +268,18 @@ describe('driver.js', function () {
...
 
     describe('#canProxy', function () {
       it('should exist', function () {
-        expect(driver.canProxy).to.be.an.instanceof(Function);
+        assert.ok(driver.canProxy instanceof Function);
       });
       it('should return true', function () {
-        expect(driver.canProxy('abc')).to.be.true;
-      });
-      it('should throw an error if session id is wrong', function () {
```
eglitise, line 285: Same question here

test/unit/session-claim-handler-specs.ts
```diff
@@ -239,9 +239,13 @@ describe('SessionClaimHandler', function () {
     await sessionClaimHandler.registerActiveSession(newDriver);
     const activeSubscriptionsBeforeClaim = mockIpc.subscriptions.filter((subscription) => subscription.isActive).length;
 
-    await expect(sessionClaimHandler.claimSessionUdid(newDriver)).to.be.rejectedWith('publish failed');
+    await assert.rejects(sessionClaimHandler.claimSessionUdid(newDriver), (err: Error) => {
+      assert.ok(err.message.includes('publish failed'));
+      return true;
```
eglitise: Same as above regarding regex matcher form

test/unit/commands/general-specs.ts
mykola-mokhnach: Done.

test/unit/commands/screenshot-specs.ts
mykola-mokhnach: Good call, that's clearer — switched to individual `assert.strictEqual` calls per field throughout the file, so a failure points straight at the mismatched property instead of dumping a whole reconstructed object.

test/unit/driver-specs.ts
mykola-mokhnach: Done.

test/unit/driver-specs.ts
mykola-mokhnach: Done.

test/unit/driver-specs.ts
mykola-mokhnach: Yes, intentionally — and it's worth flagging why. That test used chai's `.to.throw` as a bare property access (no `()`), which is a known chai footgun: unlike most chainable assertions, `.throw` only actually runs the check when invoked as a function (`.to.throw()`); accessed as a property it's a silent no-op. So this assertion had never really executed under chai — it always passed regardless of what `proxyActive()` did.

Converting to a real `assert.throws(...)` call exposed that: `proxyActive()` (and `canProxy()` below) unconditionally `return true` regardless of `sessionId` — see lib/driver.ts, which explicitly comments "we always have an active proxy to the UiAutomator2 server" / "we can always proxy to the uiautomator2 server". They never throw, by design. So the test's premise doesn't match the driver's actual (intentional) behavior, and since the old assertion was a dead no-op, not a real regression guard, I removed it rather than leave a genuinely-failing assertion or reintroduce another no-op. Happy to restore it if you'd rather it be fixed to assert the current (no-throw) behavior instead, or if `proxyActive`/`canProxy` should actually start validating `sessionId` — let me know which you'd prefer.

test/unit/driver-specs.ts
mykola-mokhnach: Same as the #proxyActive case above — the `.to.throw` (no parens) assertion was a chai no-op, and `canProxy()` also unconditionally returns `true` by design, so it never actually threw. Removed for the same reason.

test/unit/session-claim-handler-specs.ts
mykola-mokhnach: Already fixed in the same push — changed to the regex matcher form.

test/unit/driver-specs.ts
eglitise: Makes sense, then I guess it's fine to remove them.

test/unit/utils-specs.ts
```diff
@@ -126,17 +128,17 @@ describe('utils/memoize', function () {
     });
     const p1 = fn();
     const p2 = fn();
-    expect(p1).to.equal(p2);
-    expect(await p1).to.equal(7);
-    expect(calls).to.equal(1);
+    assert.strictEqual(p1, p2);
```
eglitise, line 131: Is this function comparison okay as `strictEqual` or would `deepStrictEqual` be more appropriate?

mykola-mokhnach: strictEqual is right here (and actually the more accurate one) — p1/p2 are Promise instances, and the test's whole point is verifying the memoize cache returns the *same* promise object on the second zero-arg call, i.e. reference identity, not structural equality. FWIW deepStrictEqual would behave identically for this case anyway: Node special-cases Promise instances in deepStrictEqual to only compare by reference (verified: `assert.deepStrictEqual(Promise.resolve(1), Promise.resolve(1))` throws, since they're different instances) — there's no meaningful way to structurally compare two Promises' internal state. So strictEqual is both correct and the more honest expression of intent here.

eglitise (approved):

PR #1044, opened 2026-09-03
feat: add mobile: viewportElementRect to translate web element rects to native coordinates

mykola-mokhnach:
Combines the element's CSS-pixel rect from chromedriver with window.devicePixelRatio and the WebView's own on-screen bounds (preferring the bounds Chromium reports via CDP, falling back to an XPath scan of the native view hierarchy) to compute where a web view element actually sits on the device screen. The result is clamped to the WebView's bounds, throwing an `element not interactable` error if the element is entirely offscreen once clamped.

Related to https://github.com/appium/appium/issues/4075


lib/commands/viewport.ts
```diff
@@ -47,6 +51,76 @@ export async function mobileViewPortRect(this: AndroidUiautomator2Driver): Promi
...
+  }
+
+  const [webRect, pixelRatio, webViewRect] = await Promise.all([
+    this.getElementRect(elementId),
+    getWebviewDevicePixelRatio(this),
+    getNativeWebViewRect(this),
+  ]);
+
+  const translated = {
+    x: Math.round(webViewRect.x + webRect.x * pixelRatio),
```
KazuCocoa: `getElementRect` does not return viewport-relative coordinates here. The WebDriver specification defines `x`/`y` relative to the current document element and calculates the element's absolute position: https://www.w3.org/TR/webdriver1/#get-element-rect. For example, after scrolling to `pageYOffset = 1180`, an element at document `y = 1200` is visible around viewport `y = 20`, but this calculation still translates `1200 * pixelRatio` and will usually clamp it away or throw. The test using a negative `getElementRect().y` to represent scrolling does not model actual WebDriver behavior. Please obtain viewport-relative geometry (and accumulate containing-frame offsets for elements inside iframes) before translating to native coordinates.

lib/commands/viewport.ts
```diff
@@ -47,6 +51,76 @@ export async function mobileViewPortRect(this: AndroidUiautomator2Driver): Promi
...
+    y: Math.round(webViewRect.y + webRect.y * pixelRatio),
+    width: Math.round(webRect.width * pixelRatio),
+    height: Math.round(webRect.height * pixelRatio),
```
KazuCocoa: `devicePixelRatio` alone is insufficient when the visual viewport is zoomed or panned. CSSOM View defines DPR using a visual scale factor of 1.0, while `visualViewport.scale` and its offsets are separate: https://drafts.csswg.org/cssom-view/#dom-window-devicepixelratio. After pinch zoom (or automatic input zoom), both the native origin and dimensions returned here are wrong. Could we capture the visual viewport scale/offset together with the element geometry and include them in this transform?

lib/commands/viewport.ts
```diff
@@ -76,3 +150,117 @@ export async function getDisplayDensity(this: AndroidUiautomator2Driver): Promis
...
+      return driver.uiautomator2.jwproxy.command(`/element/${elementId}/rect`, 'GET') as Promise<Rect>;
+    }),
+  );
+  const rects = settled
+    .filter((result): result is PromiseFulfilledResult<Rect> => result.status === 'fulfilled')
+    .map((result) => result.value);
+  if (!rects.length) {
+    throw new errors.NoSuchElementError('Could not determine the bounds of any native WebView element on screen');
+  }
+  return rects.reduce((largest, rect) => (rect.width * rect.height > largest.width * largest.height ? rect : largest));
```
KazuCocoa: Choosing the largest native WebView does not identify the WebView associated with the current Chromedriver target. An app may display two WebViews simultaneously while the active context belongs to the smaller one; in that case every result is translated using the other view's origin and bounds. Could this fallback correlate the candidate with the active CDP target, or fail as ambiguous instead of silently returning coordinates for an unrelated WebView?

KazuCocoa (commented):
I found three coordinate-correctness cases that should be addressed before relying on this command for native interactions.

lib/commands/viewport.ts
mykola-mokhnach: Fixed — thanks for catching this. `getElementRect`'s x/y are relative to the current document/frame, not the visible viewport. The fix subtracts `window.scrollX/scrollY` and, for elements inside iframes, accumulates each same-origin ancestor frame's own on-screen offset (via `frameElement.getBoundingClientRect()` walking up `window.parent`, stopping at the first cross-origin ancestor) before scaling. See `getWebviewGeometryContext`/`GET_WEBVIEW_GEOMETRY_CONTEXT_SCRIPT` in bf347da. The test with a negative `getElementRect().y` was replaced with a realistic model (a positive document y combined with a scroll offset).

lib/commands/viewport.ts
mykola-mokhnach: Fixed — this now reads and applies `window.visualViewport.scale`/`offsetLeft`/`offsetTop` alongside `devicePixelRatio` (same combined script as the previous reply). The layout-viewport-relative position (after scroll/frame adjustment) is scaled by `visualViewport.scale` and shifted by its pan offset before applying `devicePixelRatio`; width/height are scaled by `visualScale * pixelRatio`. See bf347da.

lib/commands/viewport.ts
mykola-mokhnach: Fixed by failing closed instead of guessing: the fallback now filters out zero-area (hidden) WebView nodes, and if more than one visible candidate remains it throws an `UnknownError` ("Found multiple visible native WebView elements...") rather than silently picking the largest. There's no reliable native-side signal to correlate a view-hierarchy node with a specific CDP target, so I didn't attempt correlation — this path is only reached when the CDP-reported bounds (which don't have this ambiguity, since they come from the specific page/target chromedriver is attached to) aren't available at all. See bf347da.

lib/commands/viewport.ts
```diff
@@ -76,3 +155,181 @@ export async function getDisplayDensity(this: AndroidUiautomator2Driver): Promis
...
+    } catch {
+      continue;
+    }
+    const {screenX: x, screenY: y, width, height, visible} = parsed;
+    if (visible === false) {
+      continue;
+    }
+    if ([x, y, width, height].every((v) => typeof v === 'number')) {
+      return {x, y, width, height} as Rect;
+    }
```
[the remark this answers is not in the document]
mykola-mokhnach: Good catch — fixed in 5ef6680. `getWebviewRectFromCdp` now requires `width > 0 && height > 0` (and also skips pages with `empty: true`) instead of just checking `typeof === 'number'`, which accepted 0. A zero-area page is now skipped in favor of the next page, or falls through to the native view-hierarchy fallback if none remain, rather than clamping everything to zero and throwing.

KazuCocoa (approved):

PR #1045, opened 2026-09-05
chore: mark the deprecated BaseDriver.getSession

mykola-mokhnach:
.

### DIFF
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -5,7 +5,6 @@ import type {
   InitialOpts,
   RouteMatcher,
   StringRecord,
-  SessionCapabilities,
 } from '@appium/types';
 import {DEFAULT_ADB_PORT, type ADB} from 'appium-adb';
 import {AndroidDriver, utils} from 'appium-android-driver';
@@ -640,13 +639,6 @@ class AndroidUiautomator2Driver
       'GET',
     )) as Partial<Uiautomator2Settings>;
     return {...driverSettings, ...serverSettings};
-  }
-
-  override async getAppiumSessionCapabilities(): Promise<SessionCapabilities<Uiautomator2Constraints>> {
-    const {capabilities} = await super.getAppiumSessionCapabilities();
-    this.log.debug('Getting session details from server to mix in');
-    const uia2Data = (await this.requireUiautomator2().jwproxy.command('/', 'GET', {})) as StringRecord;
-    return {capabilities: {...capabilities, ...uia2Data}} as SessionCapabilities<Uiautomator2Constraints>;
   }
 
   requireAdb(): ADB {
--- a/lib/driver.ts
+++ b/lib/driver.ts
@@ -4,6 +4,7 @@ import type {
   ExternalDriver,
   InitialOpts,
   RouteMatcher,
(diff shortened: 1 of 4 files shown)


lib/driver.ts
```diff
@@ -482,13 +481,6 @@ class AndroidUiautomator2Driver
     };
   }
 
-  override async getSession(): Promise<SingularSessionData<Uiautomator2Constraints>> {
```
eglitise, line 485: Isn't removing this a breaking change?

lib/driver.ts
```diff
@@ -650,9 +642,11 @@ class AndroidUiautomator2Driver
     return {...driverSettings, ...serverSettings};
   }
 
-  // needed to make the typechecker happy
   override async getAppiumSessionCapabilities(): Promise<SessionCapabilities<Uiautomator2Constraints>> {
-    return (await super.getAppiumSessionCapabilities()) as SessionCapabilities<Uiautomator2Constraints>;
+    const {capabilities} = await super.getAppiumSessionCapabilities();
+    this.log.debug('Getting session details from server to mix in');
+    const uia2Data = (await this.requireUiautomator2().jwproxy.command('/', 'GET', {})) as StringRecord;
+    return {capabilities: {...capabilities, ...uia2Data}} as SessionCapabilities<Uiautomator2Constraints>;
```
eglitise: I don't think we should mix in the server data under the `capabilities` key.
This `uia2Data` object only adds `lastScrollData`, which certainly is not a capability. If there's no other way to retrieve it, I'd suggest adding a new execute method.

mykola-mokhnach: Good point, thanks. Removed the override entirely and instead exposed the existing `getStatusBarHeight`/`getDevicePixelRatio` driver methods as new `mobile: statusBarHeight`/`mobile: pixelRatio` execute methods, alongside the already-existing `mobile: viewportRect`. That covers everything the tests needed without touching `capabilities` at all; nothing in this repo consumes `lastScrollData`, so there was no need to expose it.

mykola-mokhnach: Actually, on a closer look these are already available without any new command at all: pixelRatio, statBarHeight, and viewportRect are populated into session capabilities once at session start (in startSession via getDeviceDetails()), so they're already readable client-side off the session capabilities (driver.capabilities in a WebDriver client) with no server round trip needed. Dropped the mobile: statusBarHeight / mobile: pixelRatio execute methods I'd added and switched the functional tests to read them from driver.capabilities directly.

lib/driver.ts
mykola-mokhnach: not really, the jwp endpoint has been ditched long ago in clients. Would you prefer to make it to a breaking change instead?

eglitise: It may still be used in third-party drivers. I would suggest keeping it for now and removing it along with any other Appium 4-related breaking changes.

mykola-mokhnach: Fair point. Restored the getSession() override and marked it @deprecated with a note to remove it alongside other Appium 4-related breaking changes.

eglitise (approved):

PR #1047, opened 2026-09-12
feat: Expose webview on-screen rect via mobile: getContexts

mykola-mokhnach:
## Summary

Adds each webview's on-screen bounding rectangle to `mobile: getContexts`' output, so clients (e.g. Appium Inspector) can detect webview bounds without reimplementing their own detection heuristics ([context](https://github.com/appium/appium-inspector/pull/3155#discussion_r3962267289)).

- Each entry in the `mobile: getContexts` mapping now includes an optional `rect` (`x`, `y`, `width`, `height`, in native device screen coordinates).
- `rect` is derived the same way `mobile: viewportElementRect` already determines a webview's bounds: preferring the CDP-reported `description` bounds for each page, and falling back to a native view hierarchy scan only when there is exactly one webview and it has no usable CDP bounds (the scan can't disambiguate which native element belongs to which webview when there's more than one).
- `rect` is simply omitted when neither source yields a usable rectangle.
- Implemented by wrapping the inherited `mobileGetContexts` in the driver constructor (same pattern already used for `memoize(this.getStatusBarHeight)`), so no new command or `execute-method-map` entry is needed — `mobile: getContexts` transparently returns the enriched data.
- Also reorders `lib/commands/viewport.ts` so exported members are grouped before private helpers (no behavior change).


lib/commands/viewport.ts
```diff
@@ -156,6 +156,44 @@ export async function getWindowSize(this: AndroidUiautomator2Driver): Promise<Si
...
+export async function enrichWebviewsMappingWithRects(
+  driver: AndroidUiautomator2Driver,
+  mapping: WebviewsMapping[],
+): Promise<WebviewsMappingWithRect[]> {
+  let hasCdpRect = false;
+  for (const m of mapping as WebviewsMappingWithRect[]) {
+    for (const page of m.pages ?? []) {
+      const rect = parseWebviewRectFromPage(page as StringRecord);
+      if (rect) {
+        m.rect = rect;
```
KazuCocoa: [P2] Avoid choosing an arbitrary page's bounds for a multi-page context

A single mapping can contain multiple visible WebViews from the same process, with different rectangles in `pages`. This loop assigns the first usable rectangle to the entire context; reversing the page order changes the returned `rect`, even though the visible WebViews have not changed. A consumer such as Inspector can therefore use the bounds of a different WebView. Could we expose bounds per page, or omit the context-level rectangle when the usable page rectangles are ambiguous?

lib/commands/viewport.ts
```diff
@@ -156,6 +156,44 @@ export async function getWindowSize(this: AndroidUiautomator2Driver): Promise<Si
...
+        hasCdpRect = true;
+        break;
+      }
+    }
+  }
+  if (!hasCdpRect && mapping.length === 1) {
+    try {
+      (mapping[0] as WebviewsMappingWithRect).rect = await getNativeWebViewRectFromViewHierarchy(driver);
```
KazuCocoa, line 186: [P2] Avoid applying implicit wait to optional context geometry discovery

`getNativeWebViewRectFromViewHierarchy` calls `findElOrEls`, which uses `implicitWaitForCondition`. With one discoverable context, no CDP bounds, and no native WebView currently on screen, `mobile: getContexts` now waits for the full implicit timeout before returning (for example, an extra 30 seconds even with `waitForWebviewMs=0`). Could this best-effort lookup use a non-waiting native search? Also, the existing `viewportElementRect` path calls the wrapped `mobileGetContexts`, ignores the enriched `rect`, and falls back to the hierarchy again, so this change makes that path scan twice. Reusing the already collected rectangle would avoid the duplicate lookup.

mykola-mokhnach: Good catch on both — fixed in 2c9c467:

- Switched `getNativeWebViewRectFromViewHierarchy` to call `doFindElementOrEls` directly instead of `findElOrEls`, so this best-effort lookup no longer goes through `implicitWaitForCondition` and its retry timeout.
- `getWebviewRectFromCdp` now reuses the `rect` `mobileGetContexts` already attached (via `enrichWebviewsMappingWithRects`) instead of ignoring it, so `viewportElementRect`'s fallback no longer triggers a second native-hierarchy scan.

KazuCocoa (approved):

